PR #7, opened 2020-06-24
Add prettier check

huayunh:
Changes proposed in this Pull Request:
- Add `"prettier:check": "prettier \"src/**/**.{ts,tsx,js,jsx,json,css,scss,html}\" --check"`


huayunh:
wahoo, this is more exciting than I expected

PR #9, opened 2020-07-06
Add Action List example

daileytj:
Fixes PXBLUE 1277 & 1281.

Changes proposed in this Pull Request:
- Add action list example
- Refactor Data List example
- Fix linting errors


src/app/app.component.spec.ts
```diff
@@ -5,14 +5,14 @@ import { AppModule } from './app.module';
 
 describe('AppComponent', () => {
     beforeEach(async(() => {
-        TestBed.configureTestingModule({
+        void TestBed.configureTestingModule({
```
joebochill, line 8: This is new. I haven't seen us defining types on function calls before. Is this because of the recent changes in typescript/eslint rules?

src/app/app.component.ts
```diff
@@ -17,14 +16,15 @@ export class AppComponent {
     selected: string;
 
     constructor(
-        private router: Router,
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        private readonly router: Router,
```
joebochill, line 20: this appears to be following the naming convention rule, so why ignore?

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
+<mat-toolbar color="primary">
+    <div>Action List</div>
```
joebochill, line 2: this should probably have some sort of mat typography directive on it

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
...
+    <div fxFlex style="margin: auto;"></div>
```
joebochill, line 3: pxb-spacer?

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
...
+    <button mat-icon-button (click)="onAddItem()"><mat-icon>add</mat-icon></button>
+</mat-toolbar>
+<mat-nav-list *ngIf="data.length > 0">
+    <pxb-info-list-item
+        [hidePadding]="true"
```
joebochill, line 8: the padding on these doesn't look hidden, but that might be an issue with the ILI component...are we using the latest version?

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
...
+        *ngFor="let i of data"
+        [class.selected]="isSelected(i)"
+        (click)="onSelected(i)"
+    >
+        <span pxb-title>{{ i.name }}</span>
+        <span pxb-subtitle>{{ i.details }}</span>
+        <span pxb-right-content>
+            <button
+                data-cy="action-menu"
```
joebochill, line 17: might be able to leave these off since we aren't doing cypress testing yet, but it doesn't hurt to leave it

src/app/pages/list/action-list/action-list.component.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+        fixture = TestBed.createComponent(ActionListComponent);
+        component = fixture.componentInstance;
+        fixture.detectChanges();
+    });
+
+    it('should create', () => {
+        void expect(component).toBeTruthy();
+    });
+
+    it('createItem', async(() => {
```
emclaug2, line 25: Can you update these unit tests to read like actual sentences?

example:
it should [do what] [when]

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
...
+                <mat-icon>more_vert</mat-icon>
+            </button>
+            <mat-menu #menu="matMenu">
+                <button data-cy="remove" mat-menu-item (click)="onRemoveItem(i)">Remove</button>
+                <button data-cy="details" mat-menu-item>View Details</button>
+            </mat-menu>
+        </span>
+    </pxb-info-list-item>
+</mat-nav-list>
+<pxb-empty-state *ngIf="data.length === 0" title="No Items Found" style="margin: 16px;">
```
joebochill, line 32: Try to match the styling of the React example (centered in the page, icon, etc.). You might need to add a 
```
html, body { height: 100% }
```
to index.css

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
+<mat-toolbar color="primary">
+    <div>Action List</div>
+    <div fxFlex style="margin: auto;"></div>
+    <button mat-icon-button (click)="onAddItem()"><mat-icon>add</mat-icon></button>
```
joebochill, line 4: The react example also has a delete all button up here

src/app/pages/list/data-list/data-list.component.ts
```diff
@@ -1,45 +1,33 @@
...
+        },
+        {
+            name: 'James Monroe',
+            year: 1812,
+        },
+    ];
 
     constructor() {}
 }
-
```
emclaug2, line 19: Glad you got rid of these pipes, not sure why the data set was so complicated to begin with.

src/app/pages/list/action-list/action-list.component.scss
```diff
@@ -0,0 +1,13 @@
+@import '~@pxblue/colors/palette.scss';
+
+pxb-info-list-item {
+    cursor: pointer;
+
+    ::ng-deep &.selected mat-list-item {
```
joebochill, line 6: do we need ngdeep here or is there another class we expose for this element that we could use to target it?

src/app/pages/list/action-list/action-list.component.html
```diff
@@ -0,0 +1,37 @@
...
+</mat-toolbar>
+<mat-nav-list *ngIf="data.length > 0">
+    <pxb-info-list-item
+        [hidePadding]="true"
+        *ngFor="let i of data"
+        [class.selected]="isSelected(i)"
```
joebochill, line 10: The react example doesn't do selection behavior (just a hover state). We could pull this out to match.

src/app/app.component.spec.ts
emclaug2: This is related to the rule changes in the latest version of the eslint-parser, ^3.0.0.  This and a handful of other rules will need to be tweaked in our own eslint config.

src/app/pages/list/action-list/action-list.component.html
emclaug2: I think updating to @pxblue/angular-components ^2.0.0 will fix the padding issue.

src/app/pages/list/action-list/action-list.component.scss
daileytj: This is the only way I could style the background color right now. We should probably add a class to the InfoListItem to modify the background color.

src/app/app.component.ts
daileytj: With the new linting rules you have to use an underscore for private dependencies. It's fixed now

src/app/app.component.spec.ts
joebochill: Should we update the rules quickly so we don't have to add and then remove all of these extra types in the pattern examples?

src/app/app.component.ts
joebochill: hmm, should we follow the rule and use an underscore then?

src/app/pages/list/action-list/action-list.component.scss
joebochill: Is this still needed if we removed the select logic?

src/app/pages/list/data-list/data-list.component.html
```diff
@@ -1,10 +1,13 @@
 <mat-toolbar color="primary">
-    <div>Key Value List</div>
-    <div fxFlex style="margin: auto;"></div>
+    <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div>Data List</div>
```
joebochill, line 5: mat-title? or other mat typography directive?

joebochill:
```suggestion
    <div class="mat-title">Data List</div>
```

src/app/app.component.spec.ts
emclaug2: I'll make a quick PR in the code-standards library to account for the new rules.  I'm still approving this PR but yeah we should get these changes in before any new design patterns are made.

PR #10, opened 2020-07-08
Add multiselect list, and status list design patterns

daileytj:
Fixes PXBLUE 1289, 1293.

- Add multiselect list
- Add status list


src/app/pages/list/multiselect-list/multiselect-list.component.html
```diff
@@ -0,0 +1,40 @@
+<mat-toolbar color="primary">
+    <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title">Multiselect List</div>
+    <pxb-spacer></pxb-spacer>
+    <button mat-icon-button (click)="onAddItem()"><mat-icon>add</mat-icon></button>
+</mat-toolbar>
+<mat-nav-list>
+    <pxb-info-list-item *ngFor="let item of data">
```
joebochill, line 10: Is our pxb-info-list-item automatically doing the hover/pointer/ripple behavior? If so, we should log an issue to fix that.

src/app/pages/list/multiselect-list/multiselect-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+    <pxb-empty-state title="No Items Found" style="margin: 16px;">
+        <mat-icon pxb-empty-icon>error</mat-icon>
+        <button pxb-actions mat-raised-button color="primary" (click)="onAddItem()">
+            <mat-icon>add</mat-icon>
+            Add An Item
+        </button>
+    </pxb-empty-state>
+</div>
+<footer [class.active]="selectedItems.length > 0" class="mat-simple-snackbar" data-cy="snackbar" fxLayout="row">
+    <p style="margin-bottom: 0;">{{ selectedItems.length }} selected items</p>
```
joebochill, line 31: should this have a mat-typography class on it? Body?

src/app/pages/list/multiselect-list/multiselect-list.component.scss
```diff
@@ -0,0 +1,56 @@
+mat-nav-list {
+    padding-top: 0px;
+    mat-checkbox {
+        display: inline-flex;
+    }
```
joebochill, line 5: Is this needed now that we have InfoListItem component?

src/app/pages/list/multiselect-list/multiselect-list.component.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+mat-list-icon {
+    vertical-align: center;
+}
```
joebochill, line 10: is this needed?

src/app/pages/list/multiselect-list/multiselect-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+<div *ngIf="data.length === 0" style="height: calc(100vh - 64px);">
+    <pxb-empty-state title="No Items Found" style="margin: 16px;">
+        <mat-icon pxb-empty-icon>error</mat-icon>
+        <button pxb-actions mat-raised-button color="primary" (click)="onAddItem()">
+            <mat-icon>add</mat-icon>
+            Add An Item
+        </button>
+    </pxb-empty-state>
+</div>
+<footer [class.active]="selectedItems.length > 0" class="mat-simple-snackbar" data-cy="snackbar" fxLayout="row">
```
joebochill, line 30: Is it possible for us to use the Angular Material Snackbar for this? I did a quick look and it seems like they only support string for the action (not icons, buttons), but might be worth spending a few minutes messing around with it. On the react side we were able to switch over to using the Out-of-the-box component from MUI.

src/app/pages/list/sortable-list/sortable-list.component.html
```diff
@@ -0,0 +1,33 @@
+<mat-toolbar color="primary">
+    <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title">Sortable List</div>
+    <pxb-spacer></pxb-spacer>
+    <button mat-icon-button (click)="edit()">
+        <mat-icon>{{ draggable ? 'check' : 'edit' }}</mat-icon>
+    </button>
```
joebochill, line 9: Let's match react and go with the Edit/Save button with text instead of icon

src/app/pages/list/multiselect-list/multiselect-list.component.html
joebochill: Is this because it's in a mat-nav-list? Should we switch to mat-list if we want that off?

src/app/pages/list/sortable-list/sortable-list.component.scss
```diff
@@ -0,0 +1,42 @@
+mat-nav-list {
+    padding-top: 0px;
+}
+
+.mat-list-base .mat-list-item {
+    height: 56px;
+}
```
joebochill, line 7: Is this still needed?

src/app/pages/list/sortable-list/sortable-list.component.scss
```diff
@@ -0,0 +1,42 @@
...
+
+// the floating component when dragging
+.cdk-drag-preview {
```
joebochill, line 10: Are these still needed after switching to infolistitem? Just curious...I know we had a lot of annoying issues with the preview/placeholder styles before.

src/app/pages/list/sortable-list/sortable-list.component.html
```diff
@@ -0,0 +1,33 @@
...
+                {{ i.name }}
+            </span>
+            <span pxb-subtitle>
+                {{ i.details }}
+            </span>
+        </pxb-info-list-item>
+    </section>
+    <div cdkDropList (cdkDropListDropped)="onDrop($event)" *ngIf="draggable">
+        <pxb-info-list-item *ngFor="let i of data" cdkDrag>
+            <mat-icon pxb-icon cdkDragHandle class="handler">drag_indicator</mat-icon>
```
joebochill, line 24: Match the icon used in React

src/app/pages/list/sortable-list/sortable-list.component.html
```diff
@@ -0,0 +1,33 @@
...
+            <span pxb-title>
+                {{ i.name }}
+            </span>
+            <span pxb-subtitle>
+                {{ i.details }}
```
joebochill, line 29: In the react example, the date is shown floated to the right instead of under the name. Are we doing it this way because the drag/preview element has issues with styling if it's on the right?

src/app/pages/list/status-list/status-list.component.html
```diff
@@ -0,0 +1,17 @@
...
+    </button>
+    <div class="mat-title">Status List</div>
+    <pxb-spacer></pxb-spacer>
+</mat-toolbar>
+<mat-nav-list class="alert" *ngIf="data.length > 0">
+    <pxb-info-list-item
+        *ngFor="let item of data; let index = index"
+        [statusColor]="setAlarm(item.status) ? colors.red[500] : ''"
+    >
+        <span pxb-icon><mat-icon>home</mat-icon></span>
```
joebochill, line 13: Let's match the icon update we did for React and add some ListItemTags too

src/app/pages/list/multiselect-list/multiselect-list.component.html
daileytj: It was the mat-nav-list causing that behavior. Switching it to mat-list does the trick.

src/app/pages/list/sortable-list/sortable-list.component.scss
daileytj: We didn't need them and getting rid of them actually fixed a style issue I was seeing.

joebochill:
Things are looking good, just a couple more comments:
- It looks like we did need those cdk preview styles...compare the previous implementation (https://stackblitz.com/github/pxblue/sortable-list/tree/angular) with the current. The old one has a much smoother drag/drop animation and it's clear where the drop is going to happen.
- I like the refactoring of the 'snackbar' for multiselect into a separate component. It would be nice if it didn't re-animate each time something is selected (maybe we just need a prelimiary check to see if it's already open). If we could use the mat-snackbar here it would be nice too.
- StatusList icon colors should match the status color (it would be nice to toss a couple avatar style icons in too if possible)

emclaug2:
Getting some weird behavior on the overlay for the Drawer.

image

Also the breakpoints defined in the all the List design patterns to show the open/close menu are aligned with the breakpoints used to change the variant of the DrawerLayout.  I would recommend removing the viewport service and have the DrawerLayout use the same breakpoints as the Lists. 

image

src/app/app.component.scss
```diff
@@ -2,6 +2,7 @@
 
 ::ng-deep .pxb-blue .mat-drawer-container {
     background-color: map-get($pxb-white, 50) !important;
+    min-height: 100vh !important;
 }
```
joebochill, line 6: I would prefer to handle this by setting a rule for:
```
html, body { height: 100%; }
```
(the drawer layout is already set to fill 100% of the height of its parent, so the 100vh is kind of limiting and overkill)

and you should be able to set
```
.pxb-drawer-layout-nav-content { bakground: white; }
```
to do the background, assuming your ViewEncapsulation is configured correctly.

Side note: I don't think our drawer layout should be setting a background color at all...it should be inheriting from whatever is set on the body by default.

src/app/app.component.scss
```diff
@@ -2,6 +2,7 @@
...
 
 .right-panel {
```
joebochill, line 8: these right-panel classes are never used

src/app/pages/list/sortable-list/sortable-list.component.scss
joebochill: We still need these for the proper UX/transitions/animations

joebochill:
Outstanding issues that I see:

- Would be nice to have avatars on the status list example
- Couple styling issues mentioned above
- Drag/Drop on multiselect example is busted (as discussed yesterday)

If you want to break the multiselect work into a separate PR, I think the other two are set to go in.

src/app/app.component.scss
```diff
@@ -1,19 +1,10 @@
...
+    height: 100%;
 }
 
-@media screen and (max-width: 960px) {
-    .right-panel {
-        display: none;
-    }
+.pxb-drawer-layout-nav-content {
+    background-color: map-get($pxb-white, 50) !important;
 }
```
joebochill, line 10: these can go in styles.scss since they are global rules (and you won't need the important). 
styles.scss should also remove the import of the theme...that's handled in angular.json

src/app/app.component.scss
```diff
@@ -1,18 +1 @@
 @import '~@pxblue/colors/palette.scss';
```
joebochill, line 1: probably don't need this in here if it's not used for any styles...file eliminated!

emclaug2:
Is the snackbar on the multiselect list support to slide in the from right when it appears?

emclaug2:
Besides the snackbar being a bit weird it looks good to me.

PR #13, opened 2020-07-13
Feature/sortable list

daileytj:
Fixes PXBLUE 1292.

- Add Sortable List design pattern


jeffvg:
Are we able to get like a mouse-over on handle-bar


image

PR #14, opened 2020-07-14
Feature/complex bottom sheet

daileytj:
Fixes PXBLUE 1280.

- Add Complex Bottom Sheet design pattern example

### DIFF
--- a/src/app/pages/overlays/complex-bottom-sheet/complex-bottom-sheet.component.html
+++ b/src/app/pages/overlays/complex-bottom-sheet/complex-bottom-sheet.component.html
@@ -2,29 +2,36 @@
     <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
         <mat-icon>menu</mat-icon>
     </button>
-    <div class="mat-title">Basic Bottom Sheet</div>
+    <div>Complex Bottom Sheet</div>
     <pxb-spacer></pxb-spacer>
     <button mat-icon-button (click)="showBottomSheet()">
         <mat-icon>more_vert</mat-icon>
     </button>
 </mat-toolbar>
-<mat-list *ngIf="data.length > 0">
+<mat-list *ngIf="data && data.length > 0">
     <pxb-info-list-item
         *ngFor="let i of data; let index = index"
+        [class.active]="i.active"
         [avatar]="true"
         [statusColor]="i.active ? Colors.red[500] : ''"
     >
-        <mat-icon pxb-icon class="padded" *ngIf="!i.active">notifications</mat-icon>
+        <mat-icon pxb-icon style="padding: 8px;" *ngIf="i.type == 'alarms' && !i.active">notifications</mat-icon>
         <mat-icon
             pxb-icon
-            class="padded"
-            *ngIf="i.active"
-            [style.color]="Colors.white[50]"
+            style="padding: 8px;"
+            *ngIf="i.type == 'alarms' && i.active"
             [style.backgroundColor]="Colors.red[500]"
+            [style.color]="Colors.white[50]"
             style="box-sizing: content-box;"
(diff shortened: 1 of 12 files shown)


src/app/pages/overlays/complex-bottom-sheet/complex-bottom-sheet.component.html
```diff
@@ -0,0 +1,41 @@
...
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title">Complex Bottom Sheet</div>
+    <pxb-spacer></pxb-spacer>
+    <button mat-icon-button (click)="showBottomSheet()">
+        <mat-icon>more_vert</mat-icon>
+    </button>
+</mat-toolbar>
+<mat-list *ngIf="data && data.length > 0">
+    <pxb-info-list-item
```
joebochill, line 12: Looks like something goofy is happening with the ListItems again (might be box-sizing issue again):
Should look like:
image

Note the icons are vertically centered with the text and left aligned to the edge of the avatar

Currently looks like:

image

src/app/pages/overlays/complex-bottom-sheet/complex-bottom-sheet.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    it('should create the service', async(() => {
+        void expect(service).toBeTruthy();
+    }));
+
+    it('updateSort works', () => {
+        service.updateSort(FiltersList.EVENT_TYPE);
+        void expect(service.activeSort).toBe(FiltersList.EVENT_TYPE);
+    });
+
+    it('updateFilters works', () => {
```
joebochill, line 42: can we update these test names to match the format `it('should......')`

src/app/pages/overlays/complex-bottom-sheet/data.service.ts
```diff
@@ -0,0 +1,136 @@
+/* eslint-disable default-case */
```
joebochill, line 1: why are we disabling the default case rule?

src/app/pages/overlays/complex-bottom-sheet/data.service.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    passData(): void {
+        this.dataSource.next(this.sortAlarms(this.filterAlarms(this.fullData)));
+    }
+
+    updateSort(sortType: string): void {
+        // eslint-disable-next-line no-console
+        console.log('data: ', sortType);
```
joebochill, line 83: no console logs?

src/app/pages/overlays/complex-bottom-sheet/complex-bottom-sheet.component.html
huayunh: From a design point of view I personally like the second one better (did I do that in the React mono repo too?)

daileytj: For some reason, there was padding on the icons...

daileytj: @huayunh The react mono repo example has the icons aligned like the second picture

joebochill: The first picture is how the ILI _should_ look (that is, Eric said it). Everything is aligned on the Material keylines. Ian and I have gone back and forth on this a bunch in the past. Center aligning makes it so the icons don't align with anything else in the application. I agree it tends to look nicer in this one case where you have a mix of icons and avatars, but I would advocate for sticking with the left alignment in the majority of cases (and the default). You can achieve the center aligning (as we have done in the React example) by using a transparent avatar background.

PR #16, opened 2020-07-14
Feature/responsive table

daileytj:
Fixes PXBLUE 1290.

- Add Responsive Table design pattern example
- Update mat-toolbar to be position:sticky throughout all examples


emclaug2:
The `margin-left` in the app bar's button for each design pattern can be moved to the styles.scss file now that you have a `app-bar` class. 

Maybe something like:

`.app-bar {
    position: sticky;
    top: 0;
    z-index: 1000;
    button:first-child {
         margin-left: -8px
     }
}`

joebochill:
> The `margin-left` in the app bar's button for each design pattern can be moved to the styles.scss file now that you have a `app-bar` class.
> 
> Maybe something like:
> 
> `.app-bar { position: sticky; top: 0; z-index: 1000; button:first-child { margin-left: -8px } }`

We might want to leave it where it is to preserve modularity (copy paste the pattern folder and go)...but since that button ties in with the global drawer, I could see it going either way

PR #20, opened 2020-07-15
Feature/dynamic stepper

daileytj:
Fixes PXBLUE 1282.

- Add dynamic stepper design pattern example
- Update tsconfig files

### DIFF
--- a/src/app/pages/dynamic-stepper/dynamic-stepper.component.html
+++ b/src/app/pages/dynamic-stepper/dynamic-stepper.component.html
@@ -1,0 +1,75 @@
+<mat-toolbar color="primary" class="app-bar">
+    <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title">Dynamic Stepper</div>
+    <pxb-spacer></pxb-spacer>
+    <button
+        *ngIf="!procedureFinished"
+        mat-icon-button
+        matTooltip="Clear all steps"
+        (click)="onRemoveAll()"
+        data-cy="delete-all"
+    >
+        <mat-icon>delete</mat-icon>
+    </button>
+</mat-toolbar>
+<div *ngIf="!procedureFinished">
+    <mat-vertical-stepper [linear]="true" #stepper>
+        <mat-step *ngFor="let step of steps; let i = index" [completed]="step.completed">
+            <form>
+                <ng-template matStepLabel
+                    >{{ step.title || 'Choose an action' }}
+                    <mat-icon
+                        class="delete_icon"
+                        *ngIf="stepper.selectedIndex === i"
+                        data-cy="delete-step"
+                        (click)="removeStep(i)"
+                        matTooltip="Remove step"
+                        matTooltipPosition="right"
+                        >delete</mat-icon
+                    >
+                </ng-template>
+                <mat-radio-group class="radio-group" (change)="changeStepSelection($event, i)" data-cy="radiogroup">
+                    <mat-radio-button style="margin: 10px 0;" *ngFor="let option of stepOptions" [value]="option.value">
+                        {{ option.label }}
+                    </mat-radio-button>
+                </mat-radio-group>
+            </form>
+        </mat-step>
+        <mat-step state="add" [completed]="false" [editable]="false">
+            <ng-template matStepLabel>
+                <button
+                    mat-flat-button
+                    [disabled]="!allCompleted"
+                    data-cy="add-step"
+                    (click)="addItem()"
+                    class="add_button"
+                    [disableRipple]="!allCompleted"
+                >
+                    Add a Step
+                </button>
+            </ng-template>
+        </mat-step>
+        <ng-template matStepperIcon="edit" let-index="index">
+            {{ index + 1 }}
+        </ng-template>
+        <ng-template matStepperIcon="done" let-index="index">
+            {{ index + 1 }}
+        </ng-template>
+        <ng-template matStepperIcon="add">
+            <mat-icon>add</mat-icon>
+        </ng-template>
+    </mat-vertical-stepper>
+    <div style="padding-left: 20px;">
+        <button mat-raised-button color="primary" data-cy="done" (click)="handleDone()">
+            DONE
+        </button>
+    </div>
+</div>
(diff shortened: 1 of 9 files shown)


jeffvg:
After adding step seeing this in console

Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'tabIndex': '0'. Current value: '-1'.
    Angular 35
    RxJS 5
    Angular 9

src/app/pages/dynamic-stepper/dynamic-stepper.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+
+describe('DynamicStepperComponent', () => {
+    let component: DynamicStepperComponent;
+    let fixture: ComponentFixture<DynamicStepperComponent>;
+
+    beforeEach(async(() => {
+        void TestBed.configureTestingModule({
+            imports: [
+                DynamicStepperModule,
```
emclaug2, line 21: You should be able to replace all these individual imports with DynamicStepperModule

src/app/pages/dynamic-stepper/dynamic-stepper.component.html
```diff
@@ -0,0 +1,75 @@
...
+                        *ngIf="stepper.selectedIndex === i"
+                        data-cy="delete-step"
+                        (click)="removeStep(i)"
+                        matTooltip="Remove step"
+                        matTooltipPosition="right"
+                        >delete</mat-icon
+                    >
+                </ng-template>
+                <mat-radio-group class="radio-group" (change)="changeStepSelection($event, i)" data-cy="radiogroup">
+                    <mat-radio-button style="margin: 10px 0;" *ngFor="let option of stepOptions" [value]="option.value">
```
joebochill, line 34: Can we update all the margins/paddings in here that are using a multiple of 5 (old standard) to use multiples of 8 (material standard)?

joebochill: There might be an instance or two of this in the patterns that are already merged in too

src/app/pages/dynamic-stepper/dynamic-stepper.component.html
```diff
@@ -0,0 +1,75 @@
...
+    </mat-vertical-stepper>
+    <div style="padding-left: 20px;">
+        <button mat-raised-button color="primary" data-cy="done" (click)="handleDone()">
+            DONE
+        </button>
+    </div>
+</div>
+<div *ngIf="procedureFinished" style="padding: 20px 30px;" data-cy="reset-page">
+    <p>Procedure created successfully.</p>
+    <button mat-raised-button color="primary" (click)="handleReset()" data-cy="reset">
```
joebochill, line 72: Should we use an empty state for this?

src/app/pages/dynamic-stepper/dynamic-stepper.component.scss
```diff
@@ -0,0 +1,37 @@
...
+    padding: 16px 64px;
+    background: transparent;
+    text-align: left;
+
+    &[disabled] {
+        background: transparent;
+    }
+}
+
+::ng-deep .mat-step:last-child .mat-step-label {
```
joebochill, line 27: do we need ng deep on these (e.g., if we change the viewEncapsulation)?

src/app/pages/dynamic-stepper/dynamic-stepper.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    constructor(
+        private readonly _drawerService: StateService,
+        private readonly _breakpointObserver: BreakpointObserver,
+        private readonly _formBuilder: FormBuilder
+    ) {}
+
+    ngOnInit(): void {
+        this.RandomFormGroup = this._formBuilder.group({
+            Ctrl: ['', Validators.required],
+        });
```
joebochill, line 46: What is this random form group?

joebochill:
Should Dynamic Stepper be a top-level menu item since it's the only thing in the Steppers bucket?

emclaug2:
> Should Dynamic Stepper be a top-level menu item since it's the only thing in the Steppers bucket?

If we plan on adding more steppers then maybe we can get away with it being nested.  Otherwise the loading states is in a single item bucket too.

src/app/pages/dynamic-stepper/dynamic-stepper.component.scss
daileytj: looks like we do need ng-deep here regardless of ViewEncapsulation, unless I can find a more specific selector. Haven't had any luck yet.

src/app/pages/dynamic-stepper/dynamic-stepper.component.ts
daileytj: Removing this...

daileytj:
> > Should Dynamic Stepper be a top-level menu item since it's the only thing in the Steppers bucket?
> 
> If we plan on adding more steppers then maybe we can get away with it being nested. Otherwise the loading states is in a single item bucket too.

Let me know what you guys want me to do here

src/app/pages/dynamic-stepper/dynamic-stepper.component.html
daileytj: Will do.

joebochill:
> > > Should Dynamic Stepper be a top-level menu item since it's the only thing in the Steppers bucket?
> > 
> > 
> > If we plan on adding more steppers then maybe we can get away with it being nested. Otherwise the loading states is in a single item bucket too.
> 
> Let me know what you guys want me to do here

I was going to make the same comment on the Loading States example when we move it over. It's already a top-level item in React. And I don't suspect we're going to add any other stepper examples for a while.

src/app/pages/dynamic-stepper/dynamic-stepper.component.scss
```diff
@@ -0,0 +1,43 @@
...
+    vertical-align: middle;
+    color: map-get($pxb-black, 500);
+    font-size: 18px;
+    line-height: 24px;
+}
+
+.add-button {
+    width: 100%;
+    height: 100%;
+    padding: 16px 64px;
```
joebochill, line 18: This is going to sound very nitpicky, but the text in the Add a step row and the numbers steps are off by 4 pixels (64 vs 60). If you change this from 64px to 60, they'll line up...it'll be easier than changing the margin on the steps bc i think that's coming from theme styles.

PR #21, opened 2020-07-15
Feature/i18n

emclaug2:
- Port over i18n example
- Updated i18n menu to slide in from the right, matching react example.
- This example does not use our Drawer component for the i18n sub menu, since our component doesn't support an option to change slide direction.


src/app/pages/i18n/i18n.component.html
```diff
@@ -0,0 +1,100 @@
...
+                        {{ 'LANGUAGES.' + lang | translate }}
+                    </mat-option>
+                </mat-select>
+            </mat-form-field>
+            <button mat-raised-button color="primary" style="margin: 20px;">
+                <mat-icon>shopping_cart</mat-icon> {{ 'ADD_TO_CART' | translate }}
+            </button>
+        </mat-toolbar>
+
+        <mat-nav-list>
```
daileytj, line 75: Can you make this mat-list instead?

jeffvg:
do we want to adjust the snack-bar layout now since we have a side drawer? on a 1024 view snack-bar sitting center of entire app and not just the page...

image

src/app/pages/i18n/i18n.component.html
emclaug2: > do we want to adjust the snack-bar layout now since we have a side drawer? on a 1024 view snack-bar sitting center of entire app and not just the page...
> 
> image

Yep you're right this should be centered in the nav content, not the entire screen.  It should also be a snackbar.

joebochill: You're going to have a tough time making this use the snackbar since it only supports a single action for the right side

src/app/pages/i18n/i18n.component.html
```diff
@@ -0,0 +1,100 @@
...
+            <div>{{ 'I18N' | translate }}</div>
+            <pxb-spacer></pxb-spacer>
+            <button
+                type="button"
+                aria-label="Toggle sidenav"
+                mat-icon-button
+                (click)="openI18nMenu()"
+                style="margin-left: -4px; margin-right: 12px;"
+            >
+                <mat-icon aria-label="Side nav toggle icon">menu_open</mat-icon>
```
joebochill, line 54: We should flip this icon (scaleX -1) when we're viewing RTL languages

src/app/pages/i18n/i18n.component.html
```diff
@@ -0,0 +1,100 @@
...
+                    (ngModelChange)="translate.use(langeSelect.value)"
+                    [(ngModel)]="selectedLanguage"
+                >
+                    <mat-option *ngFor="let lang of enabledLocales; let i = index" [value]="enabledLocales[i]">
+                        {{ 'LANGUAGES.' + lang | translate }}
+                    </mat-option>
+                </mat-select>
+            </mat-form-field>
+            <button mat-raised-button color="primary" style="margin: 20px;">
+                <mat-icon>shopping_cart</mat-icon> {{ 'ADD_TO_CART' | translate }}
```
joebochill, line 71: We should probably flip this icon too (it's not really directional, but it'll be nice to have an example for teams to know how to do it)

src/app/pages/i18n/i18n.component.html
joebochill: I don't mind this one being a mat-nav-list since the click behavior is functional (checks the checkbox)

src/app/pages/i18n/i18n.component.scss
```diff
@@ -0,0 +1,112 @@
...
+        overflow-x: hidden;
+    }
+    .toolbar-header {
+        background-color: map-get($pxb-blue, 500);
+        color: white;
+        min-height: 150px;
+    }
+    .logo-row {
+        height: 72px;
+        margin-left: 10px;
```
joebochill, line 21: can we clean up the margin/padding multiples of 5 and convert to multiples of 8?

src/app/pages/i18n/i18n.component.scss
```diff
@@ -0,0 +1,112 @@
...
+
+.select-language {
+    font-size: 1rem;
+    background-color: map-get($pxb-white, 50);
+    mat-form-field {
+        color: map-get($pxb-black, 500);
+    }
+}
+
+footer {
```
joebochill, line 72: agree that this should be centered in the main content (with drawer open or closed)

emclaug2: I managed to get this centered by applying a margin-left to the snack-bar overlay container.   I originally wanted to provide a custom AppOverlayContainer that only overlaps the content found inside our DrawerLayout's content panel, but this idea didn't pan out.

I got this idea from this thread: https://github.com/angular/components/issues/7764

It worked well for the snackbar, but the snackbar is only 1 of 3 elements that use the overlay container.  The MatSelectItem and MatToolTip also use this Overlay, and when I overwrote the default AppOverlayContainer their positioning was a bit off.   Just used plain old css to fix this one.

src/app/pages/i18n/i18n.component.scss
emclaug2: Yep, multiples of 8 now

src/app/pages/i18n/i18n.component.html
emclaug2: I got a snackbar working; it was a bit of headache trying to center it relative to the i18n content.

src/app/pages/i18n/i18n.component.html
emclaug2: Added this in so the shopping cart and secondary menu icon are inverted for RTL languages now.

src/app/pages/i18n/i18n.component.html
emclaug2: Added this in so the shopping cart and secondary menu icon are inverted for RTL languages now.

huayunh:
(as a general rule of thumb, all the icons should be flipped, except for those with text in it)

src/app/pages/i18n/i18n.component.html
```diff
@@ -0,0 +1,91 @@
...
+                <mat-icon [class.rtl]="rtl">shopping_cart</mat-icon> {{ 'ADD_TO_CART' | translate }}
+            </button>
+        </mat-toolbar>
+
+        <mat-nav-list>
+            <pxb-info-list-item
+                *ngFor="let fruit of fruits"
+                (click)="toggleFruit(fruit)"
+                [chevron]="true"
+                [hidePadding]="true"
```
joebochill, line 83: maybe false for now so it looks a little more like a typical ILI

joebochill:
> (as a general rule of thumb, all the icons should be flipped, except for those with text in it)

Can we skip the ones that are horizontally symmetrical?

emclaug2:
> > (as a general rule of thumb, all the icons should be flipped, except for those with text in it)
> 
> Can we skip the ones that are horizontally symmetrical?

I just kept it non-discriminatory and added the .rtl class to all the mat-icons.  It'll keep things simpler imo.

huayunh:
I won't tell anyone 🤫

emclaug2:
> > > (as a general rule of thumb, all the icons should be flipped, except for those with text in it)
> > 
> > 
> > Can we skip the ones that are horizontally symmetrical?
> 
> I just kept it non-discriminatory and added the .rtl class to all the mat-icons. It'll keep things simpler imo.

Maybe the biggest one that matters (the chevron) is the only one that isn't inverted since it's a prop on the infoListItem right now.

src/app/pages/i18n/i18n.component.html
```diff
@@ -0,0 +1,91 @@
...
+            <pxb-spacer></pxb-spacer>
+            <button
+                [matTooltip]="'I18N_MENU_TOOLTIP' | translate"
+                matTooltipPosition="below"
+                matTooltipHideDelay="200"
+                type="button"
+                aria-label="Toggle sidenav"
+                mat-icon-button
+                (click)="toggleI18nMenu()"
+                style="margin-left: -4px; margin-right: 12px;"
```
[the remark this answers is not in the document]
emclaug2: I adjusted the styles a bit so that they're less silly.

src/app/pages/i18n/i18n.component.html
emclaug2: Yep, this'll make things easier :)

huayunh:
image

huayunh:
Annnnd it looks like a lot of the styles on Angular are disagreeing with React (colors and paddings)?

PR #23, opened 2020-07-16
Feature/loading states

emclaug2:
- Add loading states


jeffvg:
looks like the hero banner? area needs a little adjustments. Icon battery percentage and full color is off too...

image

jeffvg:
Are we planning for both empty and loading states? Angular pattern site vs React pattern site.

image

jeffvg:
placeholders seem to load up in correct area....

image

emclaug2:
> placeholders seem to load up in correct area....
> 
> image

Good catch Jeff, I've fixed the navigation hierarchy and hero the hero spacing.  Still looking into why the batteries are appearing differently now.

src/app/pages/loading-states/components/placeholder-card.component.ts
```diff
@@ -0,0 +1,32 @@
...
+                <div class="ph-row" *ngFor="let listitem of [0, 1, 2]">
+                    <div class="ph-col-12 big"></div>
+                </div>
+            </div>
+        </div>
+    `,
+    styles: [
+        `
+            .ph-item {
+                padding: 10px 0;
```
joebochill, line 23: 16?

src/app/pages/loading-states/data.ts
```diff
@@ -0,0 +1,148 @@
...
+            unit: string;
+        }>;
+    };
+};
+
+export const dataList: LoadingData[] = [
+    {
+        name: 'placeholder1',
+        data: {
+            heroValue: 0,
```
joebochill, line 20: maybe rename heroValue to scoreValue, since all of these are hero values

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
+<mat-toolbar color="primary" flexLayout="row">
+    <button *ngIf="isSmall" mat-icon-button (click)="toggleMenu()" style="margin-left: -8px;">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <h6>Loading Example</h6>
```
joebochill, line 5:
```suggestion
    <h6>Loading States</h6>
```

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+                                percent="{{ device.data.loadValue }}"
+                                size="36"
+                                pxb-primary
+                            ></pie-progress>
+                            <pxb-channel-value value="{{ device.data.loadValue }}" units="%"> </pxb-channel-value>
+                        </pxb-hero>
+                        <pxb-hero label="Battery" value="{{ device.data.battery }}" units="%">
+                            <battery-progress
+                                [style.color]="interpreter.interpretColor(device.data.battery)"
+                                percent="{{ device.data.battery }}"
```
joebochill, line 37: strange these are still showing 100% full when they load. The Pie works as expected with the battery data, so this might be a problem with the progress icon itself.

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+                                size="36"
+                                pxb-primary
+                            ></battery-progress>
+                        </pxb-hero>
+                    </pxb-hero-banner>
+                    <mat-list-item *ngFor="let list of device.data.channels">
```
joebochill, line 43: Can we use InfoListItem?

src/app/pages/loading-states/loading-states.component.scss
```diff
@@ -0,0 +1,39 @@
+@import '~@pxblue/colors/palette.scss';
+
```
joebochill, line 2: We should be able to eliminate some of these rules if we use InfoListItem

src/app/pages/loading-states/loading-states.component.scss
```diff
@@ -0,0 +1,39 @@
...
+.padding0 {
+    padding: 0;
+}
+
+mat-grid-tile {
+    overflow: visible;
+}
+
+::ng-deep .pxb-hero-banner {
```
joebochill, line 37: ngdeep plus !important? I don't like seeing these for what's probably a very common tweak that users will want to make on the hero banner.

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+<div style="padding: 20px;">
+    <mat-grid-list [cols]="cols" gutterSize="12px" rowHeight="270px">
+        <mat-grid-tile *ngFor="let device of deviceList">
+            <mat-card class="padding0">
+                <app-placeholder-card *ngIf="!device.data" id="placeholder"></app-placeholder-card>
+
+                <mat-list class="padding0" *ngIf="device.data" id="data-list">
+                    <pxb-hero-banner [divider]="true">
+                        <pxb-hero label="Grade" value="{{ device.data.heroValue }}" units="/100" size="36">
+                            <span [style.color]="interpreter.interpretGrade(device.data.heroValue)" pxb-primary>
```
joebochill, line 21: Does it make more sense to do these data manipulations as a service or as pipes, e.g.:
```
[style.color]="device.data.value | getColor"
```

emclaug2:
The battery icons seem to be not working only whenever they are placed inside of a hero-banner in version ^2.0.0 of our angular library.  I thought that this issue might have had something to do with us using different versions of angular between our original loading states demo vs our new design patterns mono repo, but I was able to reproduce this issue in our showcase app; batteries inside of heroes will always be full, but batteries outside will show as expected.

This should be a separate issue, probably opened on our component library repo.

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+
+<div style="padding: 20px;">
+    <mat-grid-list [cols]="cols" gutterSize="12px" rowHeight="270px">
+        <mat-grid-tile *ngFor="let device of deviceList">
+            <mat-card class="padding0">
+                <app-placeholder-card *ngIf="!device.data" id="placeholder"></app-placeholder-card>
+
+                <mat-list class="padding0" *ngIf="device.data" id="data-list">
+                    <pxb-hero-banner [divider]="true">
+                        <pxb-hero label="Grade" value="{{ device.data.heroValue }}" units="/100" size="36">
```
joebochill, line 20: Progress icons are not being scaled correctly in the Hero (we say 36, but the icons are still 24)...probably an issue with the hero and quite possibly the reason we didn't use progress icons in the example before.

joebochill:
> The battery icons seem to be not working only whenever they are placed inside of a hero-banner in version ^2.0.0 of our angular library. I thought that this issue might have had something to do with us using different versions of angular between our original loading states demo vs our new design patterns mono repo, but I was able to reproduce this issue in our showcase app; batteries inside of heroes will always be full, but batteries outside will show as expected.
> 
> This should be a separate issue, probably opened on our component library repo.

This is the rule in the hero that is causing the battery to render incorrectly:
image

src/app/pages/loading-states/loading-states.component.html
emclaug2: Updated this

src/app/pages/loading-states/data.ts
emclaug2: Renamed scoreValue to heroValue

src/app/pages/loading-states/components/placeholder-card.component.ts
emclaug2: Yep!

src/app/pages/loading-states/loading-states.component.html
emclaug2: Updated to use infoListItem

src/app/pages/loading-states/loading-states.component.scss
emclaug2: You're right, it did

src/app/pages/loading-states/loading-states.component.scss
emclaug2: I just replaced the !important with a higher specificity rule.  The ::ng-deep is required because this LoadingState component doesn't use ViewEncapsulation.None.

src/app/pages/loading-states/loading-states.component.html
emclaug2: This is a good idea, I've replaced the InterpretService with 3 smaller pipes used to interpret stuff.

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
+<mat-toolbar color="primary" flexLayout="row">
+    <button *ngIf="isSmall" mat-icon-button (click)="toggleMenu()" style="margin-left: -8px;">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <h6>Loading States</h6>
+    <div fxFlex></div>
+    <button mat-button color="white" (click)="refreshData()">
```
joebochill, line 7: In the other examples with toolbar actions, I believe we are using icon buttons instead of normal buttons...do we want to be consistent?

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+        <mat-icon>autorenew</mat-icon>
+    </button>
+</mat-toolbar>
+
+<div style="padding: 20px;" class="loading-states-container">
```
joebochill, line 12: should we do 16, or 24 so it's a multiple of the base unit?

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+    <mat-grid-list [cols]="cols" gutterSize="12px" rowHeight="276px">
```
joebochill, line 13: should the gutter also be bigger? 16?

src/app/pages/loading-states/loading-states.component.html
```diff
@@ -0,0 +1,54 @@
...
+        <mat-grid-tile *ngFor="let device of deviceList">
+            <mat-card class="padding0">
+                <app-placeholder-card *ngIf="!device.data" id="placeholder"></app-placeholder-card>
+
+                <mat-list class="padding0" *ngIf="device.data" id="data-list">
+                    <pxb-hero-banner [divider]="true">
+                        <pxb-hero label="Grade" value="{{ device.data.scoreValue }}" units="/100" size="36">
```
joebochill, line 20: I think the attribute is `iconSize` and its default value is already 36 :)

src/app/pages/loading-states/loading-states.component.html
joebochill: We've got a separate issue to fix this in the hero now

src/app/pages/loading-states/loading-states.component.html
emclaug2: Yep, went with 24

src/app/pages/loading-states/loading-states.component.html
emclaug2: Good catch, I made this a mat-icon-button

src/app/pages/loading-states/loading-states.component.html
emclaug2: Yep used 16px here

src/app/pages/loading-states/loading-states.component.html
emclaug2: You're right, we can get away with taking this out.

joebochill:
Changes look great. One thing that I just noticed slipped through...can we swap out the text grades for our grade icons (like we have in the showcase):
image
to
image

I think we originally did it this way because we didn't have a fix for dynamically resizing the pxb icons.

emclaug2:
> Changes look great. One thing that I just noticed slipped through...can we swap out the text grades for our grade icons (like we have in the showcase):
> image
> to
> image
> 
> I think we originally did it this way because we didn't have a fix for dynamically resizing the pxb icons.

Icon Registry is not working as expected in our design pattern repo; I thought this had something to do with our patterns repo using ng9, so I generated a new project using the pxblue cli and tried to add svg pxblue icons.  Got the same problem and opened a new issue, found here: https://github.com/pxblue/icons/issues/64

emclaug2:
Since the pxblue-icons cannot be integrated at this time, I'm going to merge this PR and open a feature request so we can come back to this enhancement.

joebochill:
> > Changes look great. One thing that I just noticed slipped through...can we swap out the text grades for our grade icons (like we have in the showcase):
> > image
> > to
> > image
> > I think we originally did it this way because we didn't have a fix for dynamically resizing the pxb icons.
> 
> Icon Registry is not working as expected in our design pattern repo; I thought this had something to do with our patterns repo using ng9, so I generated a new project using the pxblue cli and tried to add svg pxblue icons. Got the same problem and opened a new issue, found here: [pxblue/icons#64](https://github.com/pxblue/icons/issues/64)

That's really 👎

PR #26, opened 2020-07-20
Add pxblue grade icons to loading states example

### LINKED ISSUE #25: Loading States should use PX Blue Grade Icons
Use the PX Blue Grade Icons

We are using text to represent grades.  
There is an issue in the pxblue icons repo right now preventing us from using these icons; not sure how to integrate these icons in an ng9 project at the moment. 

No

#### Additional Context

Current:
image

Desired: 
image

emclaug2:
Fixes #25 

- Defaults InfoListItem color to default color
- Fixes stackblitz compilation issues (dev does not run).

#### Screenshots:
image

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -25,12 +25,7 @@
                         "aot": true,
                         "assets": [
                           "src/favicon.ico",
-                          "src/assets",
-                          {
-                            "glob": "**/*",
-                            "input": "./node_modules/@pxblue/icons-svg",
-                            "output": "./assets/@pxblue/icons-svg"
-                          }
+                          "src/assets"
                         ],
                         "styles": [
                             "src/styles.scss",
--- a/angular.json
+++ b/angular.json
@@ -23,7 +23,15 @@
                         "polyfills": "src/polyfills.ts",
                         "tsConfig": "tsconfig.app.json",
                         "aot": true,
-                        "assets": ["src/favicon.ico", "src/assets"],
+                        "assets": [
+                          "src/favicon.ico",
+                          "src/assets",
+                          {
+                            "glob": "**/*",
+                            "input": "./node_modules/@pxblue/icons-svg",
+                            "output": "./assets/@pxblue/icons-svg"
+                          }
+                        ],
                         "styles": [
                             "src/styles.scss",
                             "./node_modules/@pxblue/angular-themes/theme.scss",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,4 +30,8 @@ jobs:
           name: Unit Tests
           command: |
             yarn test:ci
+      - run:
+          name: Build
+          command: |
+            yarn build
 
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
         "@pxblue/icons": "^1.0.25",
         "@pxblue/icons-svg": "^1.0.18",
         "@pxblue/ng-progress-icons": "^1.1.4",
+        "core-js": "^3.6.5",
         "placeholder-loading": "^0.2.2",
         "rxjs": "~6.5.4",
         "tslib": "^1.10.0",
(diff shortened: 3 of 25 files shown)


emclaug2:
I couldn't get the icon-svg integration to work by doing the recommended import:

`const iconSet = require("@pxblue/icons-svg/icons.svg"); // full set`

I had to enhance the angular.json to include a reference to these icon assets, and reference the url through a plain string instead of a module: 

`const iconSet = '../assets/@pxblue/icons-svg/icons.svg';`

emclaug2:
> I couldn't get the icon-svg integration to work via the recommended import:
> 
> `const iconSet = require("@pxblue/icons-svg/icons.svg"); // full set`
> 
> I had to enhance the angular.json to include a reference to these icon assets, and reference the url through a plain string instead of a module:
> 
> `const iconSet = '../assets/@pxblue/icons-svg/icons.svg';`

jeffvg:
looks like info list item isn't using #424e54 but rather #556167

joebochill:
Does this work when the project is loaded up in CodeSandbox?

emclaug2:
> Does this work when the project is loaded up in CodeSandbox?

I tried this in stackblitz but the icons will not render because stackblitz has trouble serving static assets, even though these assets are cloned from the node_modules folder. 

https://stackblitz.com/github/pxblue/angular-design-patterns/tree/feature/pxblue-icons-loading-states

emclaug2:
> looks like info list item isn't using #424e54 but rather #556167

Good catch, there was a leftover style from the original code still in there.   It has been removed and the InfoListItem color is now back to the default.

joebochill:
If it's not going to render the icons when we run it in codesandbox, then I would prefer we just leave them as-is, with the `<span>`s instead of icons since we'll be linking out from doc-it and the readmes to CodeSandbox...everything should look correct there (or at least work).

On a side-note, it looks like the angular design patterns are not running in CodeSandbox at all:
image

emclaug2:
> If it's not going to render the icons when we run it in codesandbox, then I would prefer we just leave them as-is, with the `<span>`s instead of icons since we'll be linking out from doc-it and the readmes to CodeSandbox...everything should look correct there (or at least work).
> 
> On a side-note, it looks like the angular design patterns are not running in CodeSandbox at all:

I agree with this; if we can figure out how to get stackblitz to host static images (or if they add this ability in the future) then we can pretty easily use the grade icons again. 

This PR should still be merged because it now:

- Defaults InfoListItem color to default color
- Fixes stackblitz compilation issues (dev does not run).

joebochill:
I'm still seeing the issue running in CodeSandbox. Seems tsconfig.json has an extra comma in it, but once that's removed, I still see the bootstrapModule issue.

emclaug2:
> I'm still seeing the issue running in CodeSandbox. Seems tsconfig.json has an extra comma in it, but once that's removed, I still see the bootstrapModule issue.

I've removed the comma and tried to fix the CodeSandbox issue.  This will take more research and is a larger issue outside the scope of this PR.

joebochill:
> > I'm still seeing the issue running in CodeSandbox. Seems tsconfig.json has an extra comma in it, but once that's removed, I still see the bootstrapModule issue.
> 
> I've removed the comma and tried to fix the CodeSandbox issue, no luck. This will take more research and is a larger issue outside the scope of this PR. It works in stackblitz for now.

@daileytj this is something that we will want to fix as part of your story to update the doc-it links.

angular.json
```diff
@@ -23,7 +23,15 @@
...
-                        "assets": ["src/favicon.ico", "src/assets"],
+                        "assets": [
+                          "src/favicon.ico",
+                          "src/assets",
+                          {
+                            "glob": "**/*",
+                            "input": "./node_modules/@pxblue/icons-svg",
+                            "output": "./assets/@pxblue/icons-svg"
+                          }
+                        ],
```
joebochill, line 34: Do we still need this if we're not going to use the icons?

emclaug2: You're right; I've just removed this.

PR #29, opened 2020-07-21
Add landing page

daileytj:
Fixes PXBLUE 1426.

- Add landing page

### DIFF
--- a/src/app/pages/landing-page/landing-page.component.html
+++ b/src/app/pages/landing-page/landing-page.component.html
@@ -4,7 +4,7 @@
     </button>
     <div class="mat-title">PX Blue Design Patterns</div>
 </mat-toolbar>
-<div class="mat-typography" style="padding: 10vh 24px 40px; max-width: 616px; margin: 0 auto;">
+<div class="mat-typography" style="padding: 10vh 24px 40px; max-width: 600px; margin: 0 auto;">
     <h2 class="mat-display-3" style="margin-bottom: 24px;">
         The <span [style.color]="Colors.blue[500]">Patterns</span>.
     </h2>
--- a/src/app/pages/landing-page/landing-page.component.html
+++ b/src/app/pages/landing-page/landing-page.component.html
@@ -4,7 +4,7 @@
     </button>
     <div class="mat-title">PX Blue Design Patterns</div>
 </mat-toolbar>
-<div class="mat-typography" style="padding: 10vh 24px 40px; max-width: 664px; margin: 0 auto;">
+<div class="mat-typography" style="padding: 10vh 24px 40px; max-width: 700px; margin: 0 auto;">
     <h2 class="mat-display-3" style="margin-bottom: 24px;">
         The <span [style.color]="Colors.blue[500]">Patterns</span>.
     </h2>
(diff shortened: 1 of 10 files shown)


jeffvg:
Just noticed we have app bar on landing page and react design patterns landing page doesn't....
do we want to have both the same?

jeffvg:
at first glance, all the links are uppercase.....

image

daileytj:
> at first glance, all the links are uppercase.....
> 
> image

This is coming from the theme. It uppercases mat-button text. Is this something we want to keep doing in the theme?

emclaug2:
Strange theme behavior, I would think that we should try to match against how React is handling button text.

For the landing page, can you decrease the side padding on the buttons from 16px to 8px.
And can you add `justify-content: space-between;`  to the button link container.

emclaug2:
It also looks like there is a small top/bottom margin on the react landing page buttons.

emclaug2:
There's also a button on the React landing page that says 'Explore PX Blue Design Patterns' missing on this implementation.

daileytj:
> There's also a button on the React landing page that says 'Explore PX Blue Design Patterns' missing on this implementation.

The button is there, it opens the sidenav so its only visible under 800px.

joebochill:
> Strange theme behavior, I would think that we should try to match against how React is handling button text.
> 
> For the landing page, can you decrease the side padding on the buttons from 16px to 8px.
> And can you add `justify-content: space-between;` to the button link container.

This is interesting. Early on the Angular theme did not automatically uppercase the button labels (react did). We actually updated our react theme to support lowercase button labels (since that's part of PX Blue). I wonder if Angular Material made a change at some point in one of their recent updates.

In any event, both frameworks should support lowercase button text.

src/app/pages/landing-page/landing-page.component.html
```diff
@@ -0,0 +1,63 @@
+<mat-toolbar color="primary" class="app-bar">
```
joebochill, line 1: Can we match react and only show the toolbar at mobile scale?

src/app/pages/landing-page/landing-page.component.html
```diff
@@ -0,0 +1,63 @@
...
+    <button *ngIf="isSmall" style="margin-left: -8px;" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title">PX Blue Design Patterns</div>
+</mat-toolbar>
+<div class="mat-typography" style="padding: 10vh 24px 40px; max-width: 600px; margin: 0 auto;">
```
joebochill, line 7: the mat-typography class should already be on the app-root I believe

src/app/pages/landing-page/landing-page.component.html
```diff
@@ -0,0 +1,63 @@
...
+    <p>
+        While everyone is encouraged to interact with the design pattern demos to become familiar with the interactions
+        and behaviors, this application is primarily intended for <strong>Angular developers</strong> to provide
+        examples of how to implement these patterns in their own applications.
+    </p>
+    <button *ngIf="isSmall" mat-stroked-button color="primary" style="margin-top: 24px;" (click)="toggleMenu()">
+        Explore PX Blue Design Patterns
+    </button>
+    <hr style="margin: 24px 0;" />
+    <div fxLayout="column" fxLayoutGap="8px" fxLayout.gt-sm="row">
```
joebochill, line 26: Looks great. I would love to see the animations from the React landing page being applied for Angular too.

PR #35, opened 2020-07-27
Fix stackblitz

daileytj:
- Update imports to not use 'src/' so Stackblitz will render app


huayunh:
I tried importing it into csb and this happened: 

image

PR #39, opened 2020-08-03
Feature/i18n currency

emclaug2:
- Enhance i18n example
- Add currency format example
- Add current date example

#### Screenshots:

English:
image

French:
image

Chinese:
image

### DIFF
--- a/src/app/pages/i18n/i18n.component.html
+++ b/src/app/pages/i18n/i18n.component.html
@@ -11,6 +11,9 @@
                 <mat-toolbar-row>
                     <small style="margin-top: -24px;" id="i18n">{{ 'I18N' | translate }}</small>
                 </mat-toolbar-row>
+                <mat-toolbar-row>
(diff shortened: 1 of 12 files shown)


emclaug2:
The currency translation does not convert monetary value, it instead shows how the currency should be displayed.

emclaug2:
If there is a better place to put the date or price please let me know.

jeffvg:
do we know if the euro symbol should be displayed before the currency amount? €3.93
currently we display euro after the currency amount and has a comma rather then decimal..

joebochill:
> do we know if the euro symbol should be displayed before the currency amount? €3.93
> currently we display euro after the currency amount and has a comma rather then decimal..

Most non-english speaking countries place the Euro symbol after the amount and they use commas for decimal separator and period for thousands separator.

huayunh:
I can confirm the Chinese one looks correct to me :)

PR #43, opened 2020-08-06
Close drawer on navigation, mobile viewport

### LINKED ISSUE #41: Drawer should close on navigation / variant toggle
#### Describe the bug
Drawer is open when user transitions from permanent to temporary variant (make window skinny).
Drawer is open on page change when on mobile.

#### What is the expected behavior?
Navigating pages should dismiss the drawer.
Moving from desktop to mobile variant should not show drawer.

#### What are the steps to reproduce?
Open the app, make window skinny, observe that the menu is open on mobile viewports.

emclaug2:
Fixes #41 

- Close Drawer on Navigation
- Close Drawer by default on mobile viewports
- Close Drawer when entering mobile viewports

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,9 +1,10 @@
-import { Component, ViewEncapsulation } from '@angular/core';
+import { ChangeDetectorRef, Component, ViewEncapsulation } from '@angular/core';
 import { Router } from '@angular/router';
 import { StateService } from './services/state.service';
 import * as PXBColors from '@pxblue/colors';
 import { DrawerItem, ROUTES } from './app-routing.module';
 import { ViewportService } from './services/viewport.service';
+import { BreakpointObserver, Breakpoints, BreakpointState } from '@angular/cdk/layout';
 
 @Component({
     selector: 'app-root',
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -18,12 +19,29 @@ export class AppComponent {
...
+                if (state.matches) {
+                    this.stateService.setDrawerOpen(false);
+                    this._ref.detectChanges();
+                } else {
+                    this.stateService.setDrawerOpen(true);
+                    this._ref.detectChanges();
+                }
+            });
+    }
+
```
joebochill, line 40: I'm wondering why we need this extra logic to detect when we move into mobile view. I would expect the temporary drawer to be closed when the drawerOpen variable is false without having to set it to false again. If I take this breakpoint logic out and go from big screen to little screen i see this:
image
image

The temporary drawer is open despite `open=false`. This seems like a bug in the component where we need to handle some sort of onchange when the variant changes.

emclaug2: I've found the issue and Joe you're right, this bug is stemming from within our component library.

The DrawerLayout component is updating the internal state of the Drawer when the variant is set to persistent, and this causes the Drawer's `open` state to get out of sync with the App's `open` input variable.

I'll add a fix to the component library and see if we can get that merged in before this PR.

joebochill: Awesome!

emclaug2: New PR just opened here:

https://github.com/pxblue/angular-component-library/pull/132

This PR refactors the internal state management logic of the Drawer component so that it won't override what a user passes in as the `open` input prop.

emclaug2: When this PR goes in, the AppComponent will still need to manually specify that the Drawer `open=false` whenever the variant transitions from one type to the other.

emclaug2:
Some of the logic used to explicitly set the Drawer state on viewport changes can be removed when our component library is updated.  The component library was updated to not modify the `open` variable provided to the `Drawer`.  

The logic added in this PR is not hurting anything, but will not be required after our next ng comp lib release.

PR #58, opened 2020-09-03
Feature/overlay refactor

emclaug2:
- Use Heroes in the complex bottom sheet
- Add empty state placeholder

### DIFF
--- a/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
+++ b/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
@@ -2,6 +2,13 @@
     <div class="footer-list" data-cy="btm-sheet-sort">
         <div class="mat-h3">Sort By:</div>
         <div fxLayout="row wrap" fxLayoutAlign="center center">
+            <pxb-hero label="Time" [iconSize]="36" [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
+              <button pxb-primary mat-icon-button>
+                <mat-icon mat-list-icon>access_time</mat-icon>
+              </button>
+            </pxb-hero>
+
+
             <a [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
                 <button mat-icon-button>
                     <mat-icon mat-list-icon>access_time</mat-icon>
--- a/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
+++ b/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
@@ -1,6 +1,6 @@
 <mat-nav-list>
     <div class="footer-list" data-cy="btm-sheet-sort">
-        <h3 class="mat-overline">Sort by</h3>
+        <div class="mat-h3">Sort By:</div>
         <div fxLayout="row wrap" fxLayoutAlign="center center">
             <a [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
                 <button mat-icon-button>
@@ -18,7 +18,7 @@
     </div>
     <mat-divider></mat-divider>
     <div class="footer-list" data-cy="btm-sheet-show">
-        <h3 class="mat-overline">Show</h3>
+        <div class="mat-h3">Show:</div>
         <div fxLayout="row wrap" fxLayoutAlign="center center">
             <a
                 data-cy="active-alarms"
--- a/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
+++ b/src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
@@ -1,72 +1,60 @@
 <mat-nav-list>
     <div class="footer-list" data-cy="btm-sheet-sort">
         <div class="mat-h3">Sort By:</div>
-        <div fxLayout="row wrap" fxLayoutAlign="center center">
-            <pxb-hero label="Time" [iconSize]="36" [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
-              <button pxb-primary mat-icon-button>
-                <mat-icon mat-list-icon>access_time</mat-icon>
-              </button>
+        <pxb-hero-banner>
+            <pxb-hero label="Time" [iconSize]="36" [class.active]="activeSort == filterTypes.TIME">
+                <button pxb-primary mat-icon-button (click)="sortData(filterTypes?.TIME)">
+                    <mat-icon mat-list-icon [style.fontSize.px]="24">access_time</mat-icon>
+                </button>
             </pxb-hero>
-
-
-            <a [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
-                <button mat-icon-button>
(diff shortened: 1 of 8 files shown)


src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.scss
```diff
@@ -1,22 +1,19 @@
+.mat-h3 {
+    margin-bottom: 0;
+}
+
+.active * {
+    color: #007bc1;
```
joebochill, line 6: Can we pull this from the colors package

src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
```diff
@@ -1,65 +1,60 @@
...
+            <pxb-hero label="Time" [iconSize]="36" [class.active]="activeSort == filterTypes.TIME">
+                <button pxb-primary mat-icon-button (click)="sortData(filterTypes?.TIME)">
+                    <mat-icon mat-list-icon [style.fontSize.px]="24">access_time</mat-icon>
                 </button>
-                <span mat-line style="display: block;">Time</span>
-            </a>
-            <a [class.active]="activeSort == filterTypes.EVENT_TYPE" (click)="sortData(filterTypes.EVENT_TYPE)">
-                <button mat-icon-button>
-                    <mat-icon data-cy="btm-sheet-type-btn" mat-list-icon>info</mat-icon>
+            </pxb-hero>
```
joebochill, line 9: The whole hero should be clickable, rather than just the icon
- You'll probably want to convert those mat-icon-buttons to plain mat-icons when you make this change

src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
```diff
@@ -1,65 +1,60 @@
 <mat-nav-list>
     <div class="footer-list" data-cy="btm-sheet-sort">
-        <h3 class="mat-overline">Sort by</h3>
-        <div fxLayout="row wrap" fxLayoutAlign="center center">
-            <a [class.active]="activeSort == filterTypes.TIME" (click)="sortData(filterTypes?.TIME)">
-                <button mat-icon-button>
-                    <mat-icon mat-list-icon>access_time</mat-icon>
+        <div class="mat-h3">Sort By:</div>
```
joebochill, line 3: I think you got the labels backwards...Huayun wants us to _use_ the Overline label style in both React and Angular.

joebochill (changes requested):

src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
emclaug2: Ah, okay! This is reverted now.

src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.html
emclaug2: I've made the entire Hero clickable; no more mat-buttons.

src/app/pages/overlays/complex-bottom-sheet/bottom-sheet/bottom-sheet.scss
emclaug2: Absolutely :)

emclaug2:
Once this PR gets merged I'll update React to match.

jeffvg:
The title here is in h3 tag but is rendering in all uppercase and only noticed because the cypress test is failing on it.

image

emclaug2:
@jeffvg  I've removed the  `text-transform: uppercase` from the overline typography.

This is a comparison between the two overlay examples right now:

image

I'm planning on updating React to do the following:
1. match angular tyopgraphy
2.  Use hero components
3.  Match container padding
4.  Update bottom button to have elevation

joebochill:
Both should be using the `overline` style with uppercase transform for the labels.

@huayunh I warned you that change was going to cause some confusion lol 😆

emclaug2:
The heroes now look a bit spaced out, so the fix would be to apply a max-width to each hero.  I'm going to hold off on applying that style for now since the root class of the hero is going to be changed as a part of another PR.

jeffvg:
> @jeffvg I've removed the `text-transform: uppercase` from the overline typography.
> 
> This is a comparison between the two overlay examples right now:
> 
> React is on the left, angular is on the right.
> image
> 
> I'm planning on updating React to do the following:
> 
> 1. match angular typography
> 2. Use hero components
> 3. Match overlay container padding
> 4. Update bottom button to have elevation

did that removal of text-transform: uppercase get pushed in?

emclaug2:
> > @jeffvg I've removed the `text-transform: uppercase` from the overline typography.
> > This is a comparison between the two overlay examples right now:
> > React is on the left, angular is on the right.
> > image
> > I'm planning on updating React to do the following:
> > 
> > 1. match angular typography
> > 2. Use hero components
> > 3. Match overlay container padding
> > 4. Update bottom button to have elevation
> 
> did that removal of text-transform: uppercase get pushed in?



> > @jeffvg I've removed the `text-transform: uppercase` from the overline typography.
> > This is a comparison between the two overlay examples right now:
> > React is on the left, angular is on the right.
> > image
> > I'm planning on updating React to do the following:
> > 
> > 1. match angular typography
> > 2. Use hero components
> > 3. Match overlay container padding
> > 4. Update bottom button to have elevation
> 
> did that removal of text-transform: uppercase get pushed in?

Yep, the "Sort By" and "Show" text is now upper-cased.

joebochill:
> The heroes now look a bit spaced out, so the fix would be to apply a max-width to each hero. I'm going to hold off on applying that style for now since the root class of the hero is going to be changed as a part of another PR.

That max-width would go in the example though, not in the component, right? We don't know what kind of labels people will be using, so we should probably leave that up to them.

joebochill:
Refactor looks good. Just one question, should we be using black[500] for the hero icons (in this particular use case)?

emclaug2:
> > The heroes now look a bit spaced out, so the fix would be to apply a max-width to each hero. I'm going to hold off on applying that style for now since the root class of the hero is going to be changed as a part of another PR.
> 
> That max-width would go in the example though, not in the component, right? We don't know what kind of labels people will be using, so we should probably leave that up to them.

You're right, I didn't make the change yet because the class names were going to change soon, but I can just apply css to the hero tags instead of the classes.

emclaug2:
Here's the complex bottomsheet now:

image

emclaug2:
> Refactor looks good. Just one question, should we be using black[500] for the hero icons (in this particular use case)?

I'm not sure; right now it's using the default Hero colors which I think looks alright.

joebochill:
> > Refactor looks good. Just one question, should we be using black[500] for the hero icons (in this particular use case)?
> 
> I'm not sure; right now it's using the default Hero colors which I think looks alright.

I'm wondering because the other text in the bottomsheets is black[500]

@huayunh thoughts?

emclaug2:
@joebochill  @huayunh 
Is the elevation above the cancel button area preferred over a flat divider?   This is only present on mobile viewports. 

image

huayunh:
(1) Happy with the default hero

(2) the elevation is there so that the cancel button is pinned at the bottom, and we can scroll vertically when the content overflows. This overflow typically only occurs on a mobile device held at the landscape position.

jeffvg (approved):

joebochill (approved):

PR #62, opened 2020-09-04
List design pattern updates

daileytj:
Fixes PXBLUE-1535.

- Update checkbox size 
- Update breakpoint for responsive table
- Update drag handle size for sortable list

### DIFF
--- a/src/app/pages/list/multiselect-list/multiselect-list.component.scss
+++ b/src/app/pages/list/multiselect-list/multiselect-list.component.scss
@@ -47,3 +47,8 @@ footer.active {
         min-width: 0;
     }
 }
+
+::ng-deep .mat-checkbox-inner-container {
+    width: 18px !important;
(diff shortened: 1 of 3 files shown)


joebochill (approved):

src/app/pages/list/multiselect-list/multiselect-list.component.scss
```diff
@@ -47,3 +47,8 @@ footer.active {
         min-width: 0;
     }
 }
+
+::ng-deep .mat-checkbox-inner-container {
```
emclaug2, line 51: I'm a little hesitant to put !important here.  Can you try using slightly more specific css?  Something like this perhaps:

```<mat-list class="multi-select-list">```

```
 ::ng-deep .multi-select-list .mat-checkbox-inner-container {
    width: 18px;
    height: 18px;
}
```

PR #65, opened 2020-09-08
Start to remove hover behavior from stepper

emclaug2:
- Remove some funky hover behavior
- Having trouble with removing last hover item below. 

#### Screenshots:

On select, Add a Step should not be colored:

circleasdgasdga


src/app/pages/dynamic-stepper/dynamic-stepper.component.scss
```diff
@@ -5,39 +5,47 @@
     flex-direction: column;
 }
 
+.stepper-container {
+  ::ng-deep .mat-step-header:hover,
+  ::ng-deep .mat-focus-indicator.add-button.mat-flat-button.mat-button-base.ng-star-inserted {
+    background-color: unset;
+  }
+}
```
joebochill, line 13: Should this be in the theme?

emclaug2: I think you're right, but only if I can fix the hover error from above.  Otherwise it'll be worse than before.

emclaug2: These style fixes were migrated to this PR:   https://github.com/pxblue/themes/pull/103

When the other PR gets merged in, this branch will be updated to use a new version of @pxblue/angular-themes.

emclaug2:
Do not merge this PR in yet; it requires a new theme package to published.

joebochill:
Looks good. Would be nice if the background of the page filled at least the height of the screen:
image

src/app/pages/dynamic-stepper/dynamic-stepper.component.html
```diff
@@ -15,7 +15,7 @@
         <mat-icon>delete</mat-icon>
     </button>
 </mat-toolbar>
-<div *ngIf="!procedureFinished">
+<div *ngIf="!procedureFinished" class="stepper-container">
```
joebochill, line 18: Is this class doing anything? I don't see it referenced anywhere else

emclaug2:
> Looks good. Would be nice if the background of the page filled at least the height of the screen:
> image

This gray area is something that I'll need to fix within the component library; this is a problem with the Drawer content area not expanding 100%.

src/app/pages/dynamic-stepper/dynamic-stepper.component.html
emclaug2: This is now removed :)

joebochill (approved):

PR #67, opened 2020-09-10
Typography Change & Fix Search Bar

emclaug2:
- Use a div instead of h4

### DIFF
--- a/src/app/pages/app-bar/search-bar/search-bar.component.html
+++ b/src/app/pages/app-bar/search-bar/search-bar.component.html
@@ -35,7 +35,7 @@
                 #searchBar
                 data-cy="searchfield"
             />
-            <button mat-icon-button (click)="searchClose()" style="margin-right: 8px;" data-cy="search-close-btn">
+            <button mat-icon-button (click)="searchClose()" data-cy="search-close-btn">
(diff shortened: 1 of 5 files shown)


src/app/pages/app-bar/search-bar/search-bar.component.html
```diff
@@ -35,7 +35,7 @@
                 #searchBar
                 data-cy="searchfield"
             />
-            <button mat-icon-button (click)="searchClose()" style="margin-right: 8px;" data-cy="search-close-btn">
+            <button mat-icon-button (click)="searchClose()" data-cy="search-close-btn">
```
joebochill, line 38: Looks like we missed updating the search bar padding in the other story...we need to refactor a little bit to eliminate the inline padding of 5px on line 26. Maybe change toolbar to have the same gutters as a regular one and then have a negative margin here, or keep this change and make the toolbar 8px. I'll leave that up to you.

emclaug2: I applied the same logic here as with the other toolbars; I added negative margins to the icons so they align 16px in from the toolbar edge.

joebochill (approved):

PR #78, opened 2020-10-12
Feature/ng 10

emclaug2:
- Update dependencies to Angular 10 

View here: https://stackblitz.com/github/pxblue/angular-design-patterns/tree/feature/ng-10


emclaug2:
I'm going to make similar changes for the ionic-design-patterns.

jeffvg:
seem the i18n is flowing over to list examples when LTR is enabled

emclaug2:
I fixed the language & RTL bleed.  Whenever navigating out of the i18n example, the language and directionality is reset to english and LTR. 

There was also another issue I found; the form validation example had a style that was bleeding into the i18n example.   The fix was just to add more specify style rules so the style remains scoped to the form validation example.

jeffvg:
I know we have an issue open to get this running on IE11 but now we don't load up at all.
- couple warnings

- WARNING in C:\pxblue\angular-design-patterns\src\app\app.component.ts depends on '@pxblue/colors'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies

- WARNING in C:\pxblue\angular-design-patterns\node_modules\@pxblue\ng-progress-icons\__ivy_ngcc__\fesm2015\pxblue-ng-progress-icons.js depends on 'color'. CommonJS or AMD dependencies can cause optimization bailouts.

other than that it all looks good. Verified the 80 or so cypress tests too.

emclaug2:
Those Colors warnings are new to angular 10 so I'll make a separate issue to investigate what a fix would be.  IE11 should be working now.

jeffvg (approved):
it loads up in IE now but still has issues related to bug logged.

joebochill:
I'm not seeing an issue logged for the colors. Can we link out to it? Is this the same dependency issue we discussed last week about the angular progress icons?

emclaug2:
> I'm not seeing an issue logged for the colors. Can we link out to it? Is this the same dependency issue we discussed last week about the angular progress icons?

I've updated the angular.json to whitelist the two commonjs dependencies this app uses; this'll suppress the build warnings.    This is a band-aid for now; I think we'll have to do some work in our @pxblue/colors package to provide the missing metadata that angular is expecting, unsure of what it is at the moment.

joebochill (approved):

PR #80, opened 2020-10-23
Feature/multiselect snackbar

emclaug2:
#### Changes proposed in this Pull Request:
- Fix cypress tests in circle ci

PR #81, opened 2020-10-23
update multi-select test for changes

jeffvg:
- update test id in component and change up assert text
- 
-


emclaug2:
This PR isn't required; I've updated the GUI already in #80 so that the tests will pass.

PR #83, opened 2020-12-14
Replace info list item with empty state component for searchbar example

### LINKED ISSUE #61: Searchbar example should use EmptyState component
When there are no results in the searchbar design pattern example an EmptyState component should be used to show this.

When there are no results in the searchbar design pattern example an InfoListItem is being used to show no results.

React: See https://github.com/pxblue/react-design-patterns/issues/31
ReactNative: See https://github.com/pxblue/react-native-design-patterns/issues/17

daileytj:
Fixes #61.

- Replace info list item with empty state component for searchbar example when there are no results.

### DIFF
--- a/src/app/pages/app-bar/search-bar/search-bar.component.html
+++ b/src/app/pages/app-bar/search-bar/search-bar.component.html
@@ -48,9 +48,9 @@
         <span pxb-subtitle>{{ item.party }}</span>
         <span pxb-info>{{ item.tookOffice }}</span>
     </pxb-info-list-item>
-    <pxb-info-list-item *ngIf="(list | filter: searchText).length === 0" data-cy="list-view-no-results">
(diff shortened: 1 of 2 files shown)


emclaug2 (approved):

jeffvg:
should update test on empty state. Just remove the commented line out and add the get on pxb-empty-state content like below

it('should return no results when data does not exist', () => {
        cy.get('[data-cy=search-btn]').click()
        cy.get('[data-cy=searchfield]').type('123')
        cy.get('.pxb-empty-state-content').should('contain', 'No matching presidents')
        //cy.get('[data-cy=list-view-no-results] > .mat-list-item > .mat-list-item-content').should('contain', 'No matching presidents')
        cy.get('[data-cy=search-close-btn]').click()

    });
});

PR #99, opened 2021-06-22
Bug/fix skeleton loader

ektasawant:
Fixes PXBLUE-2160 .

- Score card skeleton height and content 
- List item count
- PXB SVG icon issue


jeffvg:
Just noticed the app bar for loading & waiting category is missing mat.primary colors

image

image

Do we know how the score card skeleton will be with 'badgeOffset'

image

ektasawant:
> Do we know how the score card skeleton will be with 'badgeOffset'

As discussed we will show same score card skeleton for different variations. @huayunh, please correct me If I am missing anything.

joebochill:
In the scorecard, the spacing should be equal between the rows on the left (first two are farther apart than the last two).
Consider making the skeleton for the row text slightly narrower so you can scoot the hero skeletons a little more to the left to be closer to their true position.

In the list items, the skeletons only need to have two lines for text (instead of three in the first list)
Divider should be the same in the skeletons and the loaded content (right now the loaded content has a full divider and the skeletons have a partial divider).

huayunh:
Joe is saying make the list items spaced tighter like this: 
image

joebochill (approved):

joebochill:
@huayunh anything else you want to see on this?

src/app/pages/loading-waiting-states/skeleton-loader/components/placeholder-score-card/placeholder-score-card.component.scss
```diff
@@ -80,7 +87,8 @@
         .footer-content {
             display: flex;
             padding: 16px;
-            color: map-get($pxb-gray, 500);
+            color: map-get($pxb-gray, 200);
+            height: 3.25rem;
         }
```
huayunh, line 92: Maybe use an info list item here and set it to `disabled` so that you wouldn't have these styles

src/app/pages/loading-waiting-states/skeleton-loader/skeleton-loader.component.html
```diff
@@ -12,10 +12,10 @@
 <section class="mat-typography" [class.small]="isSmall">
     <div class="score-card-example-section">
         <h2 class="example-title">Scorecard</h2>
-        <div class="placeholder-section" *ngIf="showPlaceHolder">
+        <div class="placeholder-section" [hidden]="!showPlaceHolder">
             <app-placeholder-score-card></app-placeholder-score-card>
         </div>
-        <div class="loaded-section" *ngIf="!showPlaceHolder">
+        <div class="loaded-section" [hidden]="showPlaceHolder">
             <mat-card>
```
huayunh, line 19: Our scorecard should not be used inside a mat-card. Now we have two layers of shadows.

src/app/pages/loading-waiting-states/skeleton-loader/skeleton-loader.component.html
```diff
@@ -101,7 +101,7 @@ <h2 class="example-title">List Items</h2>
                 </mat-nav-list>
             </mat-card>
             <h2 class="example-title"></h2>
-            <mat-card class="list-item-example-two">
+            <mat-card class="list-item-example-two" [hidden]="!showPlaceHolder">
                 <mat-nav-list [style.paddingTop.px]="0">
                     <pxb-info-list-item
                         divider="full"
```
huayunh, line 107: Line 157 of this file, there should be no divider in an info list item when it's at the end of the list

src/app/pages/loading-waiting-states/skeleton-loader/skeleton-loader.component.html
```diff
@@ -198,9 +198,6 @@ <h2 class="example-title">Hero Banner</h2>
                     <pxb-hero [label]="'Temp'" [value]="96" [units]="'C'">
```
huayunh, line 198: [units]="'°C'"

src/app/pages/loading-waiting-states/skeleton-loader/components/placeholder-list/placeholder-list.component.html
```diff
@@ -1,18 +1,22 @@
 <div class="list-items-section section-one example-section">
```
huayunh, line 1: Where is this "example-section" class used?

Same file, what is that `<h2 class="example-title"></h2>` on line 32 for?

huayunh (changes requested):
# 1
List item, right chevrons are not vertically aligned
image

# 2 
Since you have that `margin-left:2px` style applied to your AM/PM text, you can take out the space in your HTML code
image

# 3
Can we give all the list `ph-item`s a 72px height? Hero banner a 120px height?

# 4 
Tags here are too tight. Refer to https://pxblue.github.io/patterns/lists#design-specifications for spacing (should be 16px apart)
image

src/app/pages/loading-waiting-states/skeleton-loader/components/placeholder-list/placeholder-list.component.html
ektasawant: Removed unused class and empty div.

jeffvg (approved):
looks good

huayunh:
To prevent your content from shifting, the card body should have the same height.
image

You may also want to add a `padding-bottom: 8px;` to the `mat-list` so that the top and bottom padding is equal
image

huayunh:
And finally.... the "1 alarm" should be red[500] with the 600 font weight. The "0 events" should be black[500] (the default color) with the 400 font weight. Like this:
image

Everything else looks good visually

ektasawant:
> To prevent your content from shifting, the card body should have the same height.

It is difficult to add fixed height to ph-item since placeholder plugin is having different div structure. So If I add any parent div and keep all card body elements inside it. It stops working. 
Could you please review below changes?
image

joebochill:
Sizing of the placeholders looks good now.

The list item tags look a little tight next to the chevron in the list items...can we give them a little more space:
image

Don't forget about Huayun's comment about the colors of the body content in the scorecard (alarms red, 0 events black)

src/app/pages/loading-waiting-states/skeleton-loader/skeleton-loader.component.html
```diff
@@ -114,7 +111,7 @@ <h2 class="example-title"></h2>
                             >check_circle
                         </mat-icon>
                         <div class="time-section" pxb-left-content>
-                            <div>8:05 <span class="ampm">AM</span></div>
+                            <div>8:05<span class="ampm">AM</span></div>
```
joebochill, line 114: I would suggest avoiding using the classes from the skeleton library for styling our main content.

huayunh: "ampm" is not part of the placeholder skeleton library...

huayunh:
All good, just fix what Joe mentioned [above](https://github.com/pxblue/angular-design-patterns/pull/99#issuecomment-867871953).

huayunh (approved):
Looks good visually. Can you make the `ph-item` class in list items to have a border radius of 4px? right now they look like this: 
image
should be this:
image

PR #107, opened 2021-07-30
Feature/fix length passcode example

ektasawant:
Fixes PXBLUE-2299 .

- Add fix length passcode

#### To Test:
-


joebochill:
Per the discussion on Friday, I think this would be better if we use a hardcoded PASS value, rather than randomly failing on any given password. Also, the user should not be able to type more than the fixed number of characters...right now if the check fails, they can continue to type additional characters.

src/app/pages/forms-and-validation/fix-length-passcode/services/customvalidation.service.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+    let service: CustomvalidationService;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({});
+        service = TestBed.inject(CustomvalidationService);
+    });
+
+    it('should be created', () => {
+        expect(service).toBeTruthy();
+    });
```
joebochill, line 15: I feel like we should be able to do more than 'renders successfully' tests for these form examples

joebochill (commented):
Pro Tip for the future: don't branch feature branch off of an existing / in progress feature branch (always branch from dev). This PR now includes the changes from your other branch (change password), making it bigger and harder to review.

src/app/pages/forms-and-validation/fix-length-passcode/services/customvalidation.service.spec.ts
emclaug2: Yeah these could be be enhanced to include something along the lines of 'it should grayed out the description when password meets criteria'

joebochill (commented):
So are we no longer showing a fail case? I think we probably will want to show that, but just have a set password that will pass (e.g., 123456).

Visually things look good.

ektasawant:
> So are we no longer showing a fail case? I think we probably will want to show that, but just have a set password that will pass (e.g., 123456).
> 
> Visually things look good.

Please review:
image

src/app/pages/forms-and-validation/fix-length-passcode/fix-length-passcode.component.html
```diff
@@ -0,0 +1,67 @@
+<mat-toolbar color="primary" class="app-bar mat-elevation-z2" data-cy="pxb-toolbar">
+    <button data-cy="toolbar-menu" *ngIf="isSmall" style="margin-left: -8px" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">Fixed Length Passcode</div>
+</mat-toolbar>
+<section class="mat-typography" [class.small]="isSmall">
+    <div class="mat-body-1 passcode-form-message">
+        <p>
+            Please enter the <strong>six-digit passcode</strong> we just send to you. The passcode is valid for 15
```
joebochill, line 10:
```suggestion
            Please enter the <strong>six-digit passcode</strong> we just sent to you. The passcode is valid for 15
```

src/app/pages/forms-and-validation/fix-length-passcode/fix-length-passcode.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+        void TestBed.configureTestingModule({
+            imports: [FixLengthPasscodeModule],
+        }).compileComponents();
+    }));
+
+    beforeEach(async () => {
+        await TestBed.configureTestingModule({
+            declarations: [FixLengthPasscodeComponent],
+        }).compileComponents();
+    });
```
joebochill, line 21: Can these be combined?

joebochill (approved):
Looks good to me

jeffvg:
looks good. We talked about the example running on mobile browser with inputmode="numeric". Was that a requirement for this? Currently on mobile web we can type letters

ektasawant:
> inputmode="numeric"

Number keypad should display in mobile since input field is accepting only numbers.
image

jeffvg:
I have latest and not seeing inputmode="numeric"

image


image

jeffvg (approved):
looks good

PR #109, opened 2021-08-04
Feature/verify on submit example

ektasawant:
Fixes PXBLUE-2321 .

- Verify on submit example
- Mobile view
- test cases

#### To Test:
- Local test URL: http://localhost:4200/forms-and-validation/verify-on-submit

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -20,6 +20,7 @@ import { SpinnerOverlaysComponent } from './pages/loading-waiting-states/spinner
 import { ContextualSpinnersComponent } from './pages/loading-waiting-states/contextual-spinners/contextual-spinners.component';
 import { SkeletonLoaderComponent } from './pages/loading-waiting-states/skeleton-loader/skeleton-loader.component';
 import { PasswordComponent } from './pages/forms-and-validation/password/password.component';
+import { VerifyOnSubmitComponent } from './pages/forms-and-validation/verify-on-submit/verify-on-submit.component';
 
 export type DrawerItem = {
     title: string;
@@ -59,6 +60,11 @@ export const ROUTES: DrawerItem[] = [
                 title: 'Password',
                 path: 'password',
                 component: PasswordComponent,
+            },
+            {
+                title: 'Verify on Submit',
+                path: 'verify-on-submit',
+                component: VerifyOnSubmitComponent,
             },
         ],
     },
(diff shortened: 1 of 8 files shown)


daileytj:
Be sure to follow the design here.

This example should:
[ ] Have some top level information on how this example works
[ ] Have a numeric only field
[ ] Have a limit of 6 characters
[ ] Have a reset form button instead of search device button
[ ]  Have an outlined button instead of a flat button
[ ] Automatically submit when there are 6 characters
[ ] Show an error message on blur 
[ ] Have a loading spinner in the form field during submission
[ ] Have a passing case of "123456" that shows faded content and a green check

Screen Shot 2021-08-04 at 7 15 21 AM

joebochill (commented):
Looks pretty good. One thing I would expect is if there is an error, it should clear when the form is resubmitted. E.g.,: You search for 12345 and get an error, you don't change the value and search again (maybe it was a network error or connectivity error of some kind and the value is correct as is)

joebochill:
> Be sure to follow the design here.

@daileytj , I think you are looking at the wrong design file...this PR is for the verifyOnSubmit, not the fixed length passcode:
image

daileytj:
> > Be sure to follow the design here.
> 
> @daileytj , I think you are looking at the wrong design file...this PR is for the verifyOnSubmit, not the fixed length passcode (see #107):

Ahh, you're right. Sorry @ektasawant, ignore my previous comment.

ektasawant:
> Looks pretty good. One thing I would expect is if there is an error, it should clear when the form is resubmitted. E.g.,: You search for 12345 and get an error, you don't change the value and search again (maybe it was a network error or connectivity error of some kind and the value is correct as is)

Do I need to show hint as below to let user know how to check valid state.
image

ektasawant:
> maybe I missed this but do we have a success state for this

I have updated my PR. User need to resubmit form with same input to check valid state. Please check https://github.com/pxblue/angular-design-patterns/pull/109#pullrequestreview-722165382

joebochill:
From the pattern description:
image
, so it seems there is no 'success screen' expected for the form itself.

If we want to show success, I would suggest we replace the form entirely with something like an empty state success screen "Device was added successfully" with a button to go back / reset. @huayunh do you want to weigh in on this?

I also find the "Find Device" and "Search Device" labels to be a bit confusing because this isn't really traditional search. From the pattern description, I think it would make more sense to be something like "Connect a Device" and "Add Device" (i.e., adding some sort of networked device by its serial number and then it's either not found or it's added successfully). We might also want to do what we did for the fixed length example and have a fixed number that works and everything else fails.

huayunh:
> If we want to show success, I would suggest we replace the form entirely with something like an empty state success screen "Device was added successfully" with a button to go back / reset. @huayunh do you want to weigh in on this?

Yeah we can do that. I've updated Figma for this. The new update includes some screen transition animations, so feel free to open a new PR.

ektasawant:
Adding Label as a Draft (New design changes https://www.figma.com/file/ZzZJSoGjeKdrYct57EMPS0/Design-Patterns-Mono-Repo-Redesign?node-id=570%3A122584)

PR #113, opened 2021-08-12
set errors to null while resubmitting form

### LINKED ISSUE #112: Verify on Submit - Clear previous error on search
This issue is for the Verify on Submit example. 

If you type in a value and get the error text and then immediately search again without changing the value, I would expect the error message to be cleared when the submit button is pressed and then reappear when the new error is registered (rather than the error message staying the entire time).

Error is not cleared on subsequent search. 

No

ektasawant:
Fixes #112 - Clear error while resubmitting form

- Clear while submitting form 

#### To Test:
- Type code other than `123` for e.g `1234` click on submit. 
- After getting error resubmit form without changing code.
- Current behavior: Showing previous error when you click on submit.
- Expected behaviour: When the submit button is pressed then new error should reappear

### DIFF
--- a/src/app/pages/forms-and-validation/verify-on-submit/verify-on-submit.component.ts
+++ b/src/app/pages/forms-and-validation/verify-on-submit/verify-on-submit.component.ts
@@ -56,6 +56,10 @@ export class VerifyOnSubmitComponent implements OnInit {
 
     searchDevice(): void {
         this.formSubmit = true;
+        if (this.searchDeviceForm.invalid) {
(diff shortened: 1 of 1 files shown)


daileytj:
I updated your initial PR comment to include `fixes #112`. This will automatically link to the issue and close it when the PR is merged. In the future, when fixing bugs, be sure to link the assigned bug so we know what you're fixing, and to save us some time manually closing bugs.

daileytj:
I also make it a practice to assign myself to any bug I'm working on so that another member of the team doesn't start working on it at the same time.

daileytj (approved):

PR #114, opened 2021-08-13
Feature/phone number format example

ektasawant:
Fixes PXBLUE-2321 (Phone number format) [Figma URL](https://www.figma.com/file/ZzZJSoGjeKdrYct57EMPS0/Design-Patterns-Mono-Repo-Redesign?node-id=536%3A98570)

[X] Show default US country code with placeholder.
[X] Show placeholder as per country selection.
[X] Separating phone number with space as per country phone number format.
[X] Using libphonenumber-js [libphonenumber](https://www.npmjs.com/package/libphonenumber-js) for phone number validation.
[X] On blur check phone number validation.
[X] If user enters phone number format for US and then change country, validate input again.
[X] Users will be only allowed to put in numbers. 
[X] Ignore extra digits if number of digits exceed the specified format.

[ ] Need to fix if phone number starts with `444 ...` which is invalid for US allows more than 10 digits.
[ ] Styling of `mat-select` to match design

#### Screenshots:
image
image

#### To Test:
- Local test URL: http://localhost:4200/forms-and-validation/phone-number-format


emclaug2:
Can you add maxLength requirements for the canadian phone number input? I was able to type more digits than were shown the preview and lost my formatting.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,118 @@
...
+    initForm(): void {
+        this.validatePhoneNumberForm = this._formBuilder.group({
+            phone: ['', [Validators.required, this.validatePhoneNumberInput.bind(this)]],
+            selectedCountry: ['US', Validators.required],
+        });
+    }
+
+    validatePhoneNumberInput(phoneNumberInput: AbstractControl): any {
+        const inputValue: string = phoneNumberInput.value.toString();
+        const phoneNumber: any = parsePhoneNumberFromString(inputValue, this.selectedCountry);
```
joebochill, line 56: can we avoid using the any type here? This is a prime example of a place where knowing the type is important...I have no idea what the output of the parsePhoneNumberFromString function is otherwise.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,118 @@
...
+import { parsePhoneNumberFromString, AsYouType } from 'libphonenumber-js';
+
+@Component({
+    selector: 'app-phone-number-format',
+    templateUrl: './phone-number-format.component.html',
+    styleUrls: ['./phone-number-format.component.scss'],
+})
+export class PhoneNumberFormatComponent implements OnInit {
+    isSmall: boolean;
+    selectedPhoneNumber: any;
```
joebochill, line 14: Too much use of the `any` type throughout. Using any like this kind of defeats the point of using Typescript in the first place.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,118 @@
...
+            selectedCountry: ['US', Validators.required],
+        });
+    }
+
+    validatePhoneNumberInput(phoneNumberInput: AbstractControl): any {
+        const inputValue: string = phoneNumberInput.value.toString();
+        const phoneNumber: any = parsePhoneNumberFromString(inputValue, this.selectedCountry);
+        if (phoneNumber) {
+            if (phoneNumber.isValid()) {
+                this.validPhoneNumber = this.validatePhoneNumberForm.controls['phone'].value.length;
```
joebochill, line 59: This seems like a poorly named variable. We're storing the length of an input value in a variable called 'validPhoneNumber'...that's kind of confusing to me.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
joebochill: Variable naming confusion: What is 'selectedPhoneNumber'? I don't see anywhere in the form where I am selecting a phone number.

joebochill (commented):
The auto formatting here seems really broken.

With nothing entered, the placeholder tells me it's looking for a format like ### ### ####
image

I type two digits and they are formatted ## as expected:
image

I type a third digit, and now a space is inserted:
image

I get to five digits and the space is completely removed:
image

If I type different digits, I get different formatting:
image
image

There seems to be no formatting at all happening for RU:
image


Lots of weird validation issues when you get towards the character limit:

- It lets me type 11 characters if they are all 1s
    - If I highlight the entire input and hit delete, only the last character is deleted
    - If I highlight the entire input and type a character to replace, nothing happens

ektasawant:
Closing PR since working on phone number format and dropdown styling.

src/app/app-routing.module.ts
```diff
@@ -78,6 +79,11 @@ export const ROUTES: DrawerItem[] = [
                 path: 'sectioned-form',
                 component: SectionedFormComponent,
             },
+            {
+                title: 'Phone Number Format',
+                path: 'phone-number-format',
+                component: PhoneNumberFormatComponent,
+            },
```
daileytj, line 86: Can we move this up so that the DrawerNavItems are in the same order as the designs?

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
```diff
@@ -0,0 +1,42 @@
+@import '~@pxblue/colors/palette.scss';
+
+section.mat-typography {
+    padding-top: 40px;
+    max-width: 480px;
+    margin: 0 auto;
+    .phone-number-form {
+        display: flex;
+        .country-code-select {
+            max-width: 112px;
```
daileytj, line 10:
```suggestion
            max-width: 135px;
```

Let's give that select arrow some breathing room. It looks like you set the min-width to match the desktop select input design that said "area code". I'd recommend changing it to match the select input width that is in the mobile design.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
```diff
@@ -0,0 +1,42 @@
...
+
+::ng-deep .mat-form-field.mat-focused.mat-primary .mat-select-arrow {
+    border-top: 0;
+    border-bottom: 5px solid;
+}
+
+::ng-deep .cdk-overlay-container .cdk-overlay-pane .selectPanel {
+    min-width: 112px !important;
+    border-bottom-right-radius: 4px;
+    border-bottom-left-radius: 4px;
```
daileytj, line 31: What are these doing?

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
```diff
@@ -0,0 +1,42 @@
...
+    background-color: map-get($pxb-white, 50);
+}
+
+::ng-deep .mat-form-field.mat-focused.mat-primary .mat-select-arrow {
+    border-top: 0;
+    border-bottom: 5px solid;
+}
+
+::ng-deep .cdk-overlay-container .cdk-overlay-pane .selectPanel {
+    min-width: 112px !important;
```
daileytj, line 29: Update this to match the updated select input width
```suggestion
    min-width: 135px !important;
```
It would be ideal to increase specificity here to avoid the `!important` flag as well.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
daileytj: These styles work with `!important`, but it would be nice to get them working without that.

src/app/pages/forms-and-validation/phone-number-format/shared/format-phone-number.pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+        let phoneNumber = value.replace(/\s/g, '');
+        switch (country) {
+            case 'US':
+            case 'CA': {
+                if (phoneNumber.length > 3 && phoneNumber.length <= 6)
+                    phoneNumber = `${phoneNumber.slice(0, 3)} ${phoneNumber.slice(3)}`;
+                else if (phoneNumber.length > 6)
+                    phoneNumber = `${phoneNumber.slice(0, 3)} ${phoneNumber.slice(3, 6)} ${phoneNumber.slice(6)}`;
+                return phoneNumber;
+            }
```
daileytj, line 20: Let's clean this up a bit by moving these cases down with the default case since it's the same.

src/app/pages/forms-and-validation/phone-number-format/shared/format-phone-number.pipe.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { FormatPhoneNumberPipe } from './format-phone-number.pipe';
+
+describe('FormatPhoneNumberPipe', () => {
+    it('create an instance', () => {
```
daileytj, line 4: Can we add some tests here to verify output when using the pipe?

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,120 @@
...
+        this.validatePhoneNumberForm.controls.phone.markAsUntouched();
+    }
+
+    allowNumbersOnly(event: KeyboardEvent): void {
+        const allowedRegex = /[0-9]/g;
+        if (!event.key.match(allowedRegex)) {
+            event.preventDefault();
+        }
+        this.countryChange = false;
+    }
```
daileytj, line 100: Is there a reason you are using this method instead of using the `matInput type="number"`?
https://material.angular.io/components/input/overview#supported-input-types

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,120 @@
...
+            selectedCountry: ['US', Validators.required],
+        });
+    }
+
+    checkPhoneNumberPattern(phoneNumber: string, countryCode: string): boolean {
+        switch (countryCode) {
+            case 'US':
+            case 'CA': {
+                return /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/.test(phoneNumber);
+            }
```
daileytj, line 63: Move these down with the default case.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.html
```diff
@@ -0,0 +1,47 @@
...
+                inputmode="numeric"
+                [maxlength]="selectedCountry['maxLength']"
+                (keypress)="allowNumbersOnly($event)"
+                (focus)="removeError()"
+                [placeholder]="selectedCountry['placeholder']"
+            />
+
+            <mat-error
+                *ngIf="
+                    validatePhoneNumberForm.get('phone').hasError('phoneNumberNotMatch') ||
```
daileytj, line 39: can we change this error key to something like "invalidFormat" or even as explicit as "invalidPhoneNumberFormat"?

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
ektasawant: I tried with input type="number", it shows arrows which are not matching our design so decided to use type="text". 
image

If we want type-"number" then I will try to hide arrow using CSS.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
ektasawant: It didn't work without important flag, since mat-select-panel calculates width and applies on the fly. 
image

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.scss
ektasawant: Removed it.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
daileytj: @huayunh Any thoughts on this `type="number"` arrow dilemma? I don't mind leaving it as a text input with the custom function, just feels like overkill when there is a number type option.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,116 @@
...
+            case 'EG': {
+                return /^\(?([0-9]{1})\)?[-. ]?([0-9]{7})$/.test(phoneNumber);
+            }
+            case 'IN': {
+                return /^\(?([0-9]{4})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{3})$/.test(phoneNumber);
+            }
+            default: {
+                return /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/.test(phoneNumber);
+            }
```
daileytj, line 71:
```suggestion
            case 'US':
            case 'CA':
            default: {
                return /^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$/.test(phoneNumber);
            }
```

I would still explicitly set a case for US and CA, but they can be grouped with the default since they were returning the same thing.

src/app/pages/forms-and-validation/phone-number-format/shared/format-phone-number.pipe.ts
daileytj: I was hoping to remove lines 15-20 and adding the `case 'US': case 'CA':` just above the `default:` like I explained in the other switch example.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.html
```diff
@@ -0,0 +1,47 @@
+<mat-toolbar color="primary" class="app-bar mat-elevation-z4" data-cy="pxb-toolbar">
+    <button data-cy="toolbar-menu" *ngIf="isSmall" style="margin-left: -8px" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">Phone Number Format</div>
+</mat-toolbar>
+<section class="mat-typography">
```
daileytj, line 7: Why is this class called "mat-typography"? It doesn't seem appropriately descriptive of what this element is.

src/app/pages/forms-and-validation/phone-number-format/phone-number-format.component.ts
```diff
@@ -0,0 +1,118 @@
...
+    selector: 'app-phone-number-format',
+    templateUrl: './phone-number-format.component.html',
+    styleUrls: ['./phone-number-format.component.scss'],
+})
+export class PhoneNumberFormatComponent implements OnInit {
+    isSmall: boolean;
+    validatePhoneNumberForm: FormGroup;
+    countries: CountryDetails[] = [
+        { code: 'US', name: '+1 (US)', placeholder: '### ### ####', maxLength: '12', errorCode: 'U.S.' },
+        { code: 'CA', name: '+1 (CA)', placeholder: '### ### ####', maxLength: '12', errorCode: 'Candian' },
```
daileytj, line 23:
```suggestion
        { code: 'CA', name: '+1 (CA)', placeholder: '### ### ####', maxLength: '12', errorCode: 'Canadian' },
```

daileytj:
This is looking really good. The only thing that looks like an issue at this point is when you change the country code after adding a phone number. I would expect the formatting to happen again with the new format with regards to the country selected and then re-validate the input. For example, I add "111 222 3333" and change the country to "RU", I would then expect an immediate input change to "111 222 33 33", but it's currently keeping the format and showing no error. If I switch to "EG" in this scenario, I get an error and no updated formatting.  I've seen this work from one country back to US/Default, but the behavior isn't consistent. It may be worthwhile to manually use your custom formatting pipe any time the country code changes to change the value of the input.

PR #117, opened 2021-08-20
Feature/list validation example

ektasawant:
Fixes PXBLUE-2322 .
Figma Link: https://www.figma.com/file/ZzZJSoGjeKdrYct57EMPS0/Design-Patterns-Mono-Repo-Redesign?node-id=536%3A98570

- Desktop view for list validation
- Mobile view 

#### Screenshots:
Desktop View:
image

Mobile View:
image

#### To Test:
- Local Test URL: http://localhost:4200/forms-and-validation/in-a-list

### DIFF
--- a/src/app/pages/forms-and-validation/in-a-list/in-a-list.component.html
+++ b/src/app/pages/forms-and-validation/in-a-list/in-a-list.component.html
@@ -1,0 +1,32 @@
+<mat-toolbar color="primary" class="app-bar mat-elevation-z4" data-cy="pxb-toolbar">
+    <button data-cy="toolbar-menu" *ngIf="isSmall" style="margin-left: -8px" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">In a List</div>
+</mat-toolbar>
+<section style="padding-top: 24px;max-width: 686px;
+margin: 0 auto;">
+    <mat-card style="padding: 0;">
+        <mat-card-content>
+            <mat-list style="padding: 0;">
+                <pxb-info-list-item  divider="partial" class="test" iconAlign="left">
+                    <div pxb-title>IP Address</div>
+                    <mat-icon pxb-icon>dns</mat-icon>
+                    <mat-form-field appearance="fill" pxb-right-content>
+                        <mat-label>Address</mat-label>
+                        <input matInput type="text" data-cy="ip-address" />
+                    </mat-form-field>
+                </pxb-info-list-item>
+    
+                <pxb-info-list-item  class="test" iconAlign="left">
+                    <div pxb-title>Insight Report</div>
+                    <div pxb-subtitle>Auto-report every 2 months</div>
+                    <mat-icon pxb-icon>timeline</mat-icon>
+                    <div class="toggle" pxb-right-content>
+                        <mat-slide-toggle></mat-slide-toggle>
+                    </div>
+                </pxb-info-list-item>
+            </mat-list>
+        </mat-card-content>
+    </mat-card>
+</section>
--- a/src/app/pages/forms-and-validation/in-a-list/in-a-list.component.html
+++ b/src/app/pages/forms-and-validation/in-a-list/in-a-list.component.html
@@ -4,29 +4,33 @@
     </button>
     <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">In a List</div>
 </mat-toolbar>
-<section style="padding-top: 24px;max-width: 686px;
-margin: 0 auto;">
-    <mat-card style="padding: 0;">
+<section class="mat-typography">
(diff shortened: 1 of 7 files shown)


emclaug2:
@huayunh What kind of input should we be using here?  Is there any animation expected whenever the user focuses or blurs the input?  

Reason why I ask is because the React version of this pattern uses a TextField with a `filled` variant but this example doesn't use the ng material components, so there's no animation.

src/app/pages/forms-and-validation/in-a-list/in-a-list.module.ts
```diff
@@ -0,0 +1,43 @@
...
+        BrowserAnimationsModule,
+        CommonModule,
+        MatToolbarModule,
+        MatIconModule,
+        MatButtonModule,
+        MatInputModule,
+        MatFormFieldModule,
+        MatDividerModule,
+        MatSelectModule,
+        MatTooltipModule,
```
emclaug2, line 36: Can you clean up this module to remove some of this things not used?

- MatButtonModule
- MatToolTipModule
- MatSelectModule
- MatCheckboxModule

these are a couple that stand out

src/app/pages/forms-and-validation/in-a-list/in-a-list.component.scss
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+.ip-address {
+    border: 0;
+    background-color: map-get($pxb-white, 200);
+    height: 40px;
+    outline: 0;
+    font-size: 16px;
+    padding: 8px 16px;
+    border-bottom: 1px solid map-get($pxb-gray, 100);
```
emclaug2, line 25: Can we use the `divider` prop on the InfoListItem here?

src/app/pages/forms-and-validation/in-a-list/in-a-list.component.html
```diff
@@ -0,0 +1,36 @@
+<mat-toolbar color="primary" class="app-bar mat-elevation-z4" data-cy="pxb-toolbar">
+    <button data-cy="toolbar-menu" *ngIf="isSmall" style="margin-left: -8px" mat-icon-button (click)="toggleMenu()">
+        <mat-icon>menu</mat-icon>
+    </button>
+    <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">In a List</div>
+</mat-toolbar>
+<section class="mat-typography">
```
daileytj, line 7: I'd like to see a different class name here since "mat-typography" is not descriptive of what this section is. I've seen this in a few places, probably just a copy/paste issue. Let's keep an eye out for this in the future so we don't keep adding this class

src/app/pages/forms-and-validation/in-a-list/in-a-list.component.scss
```diff
@@ -0,0 +1,38 @@
+@import '~@pxblue/colors/palette.scss';
+
+::ng-deep .pxb-drawer-layout-nav-content {
+    background-color: map-get($pxb-white, 50);
+}
+section.mat-typography {
+    padding-top: 24px;
+    max-width: 686px;
```
daileytj, line 8: I've been seeing a lot of these max-width settings to arbitrary numbers. I know the design shows 686px, but it would be nice to at least have this be a multiple of 8px or possibly come up with a standard for how wide we want the content to be so it's not a different value for each example. 

@huayunh is there a reasonable max-width standard we can have for these examples on desktop?

If Huayun is cool with this feel free to ignore this comment.

src/app/pages/forms-and-validation/in-a-list/in-a-list.component.scss
ektasawant: This is for showing border below input field. 
image

Not sure how to use divider prop.

daileytj: There is a `divider` prop you can add to the `pxb-info-list-item` like so
```
<pxb-info-list-item divider="partial">...
```

here's an example from our storybook
https://pxblue-components.github.io/angular/?path=/story/components-info-list-item--within-a-full-list

emclaug2: I misunderstood earlier; that divider I was referring to wasn't for the InfoListItem, it was used in the input field's bottom border.

emclaug2:
Can you remove this top-padding on mobile viewports? 
image

ektasawant:
> @huayunh What kind of input should we be using here? Is there any animation expected whenever the user focuses or blurs the input?
> 
> Reason why I ask is because the React version of this pattern uses a TextField with a `filled` variant but this example doesn't use the ng material component, so there's no animation.

@emclaug2 , I have replace it with ng-material input. Could you please review it ?

ektasawant:
> Can you remove this top-padding on mobile viewports?
> image

Fixed.

emclaug2 (approved):

emclaug2:
Looking good!

Issue #121, opened 2021-09-07
Pages missing mat toolbar row and toggle menu missing


jeffvg:
#### Describe the bug
Both the fixed length passcode and verify on submit pages toggleMenu partially implemented and not working on small views.
#### What is the expected behavior?
Resizing both pages to toggle drawer menu icon for pages
#### What are the steps to reproduce?
1. Go to 
http://localhost:4200/forms-and-validation/fixed-length-passcode
OR
http://localhost:4200/forms-and-validation/verify-on-submit
2. Resize page to any small view
3. toggleMenu not displayed

PR #123, opened 2021-09-08
fix toggle icon should display when resize to mobile size

ektasawant:
Fixes [121](https://github.com/pxblue/angular-design-patterns/issues/121) .


emclaug2 (approved):
Looking good!

Issue #104, opened 2021-07-15
Spinner overlay design should match react design pattern


ektasawant:
#### Describe the bug
Match NG design pattern with react design pattern:

1. Width should be as per design.
2. Header and footer should not scroll only card content is scrollable.
3. In mobile remove padding from body.

#### What is the expected behavior?
1. Width should be as per design.
2. Header and footer should not scroll only card content is scrollable.
3. In mobile remove padding from body.

#### Screenshots or links to minimum reproduction example
![image](https://user-images.githubusercontent.com/10433274/125732460-0917db35-7895-43d0-8c1b-13925e2c5342.png)

PR #126, opened 2021-09-09
Fix/spinner width

ektasawant:
Fixes [104](https://github.com/pxblue/angular-design-patterns/issues/104) .
Match Ng design pattern with react design (spinner overlay)
1. Width should be as per design.
2. Header and footer should not scroll only card content is scrollable.
3. In mobile remove padding from body


emclaug2 (commented):
I'm taking a look at this now but initial impressions are that there's a lot of code changes for such a small issue.  What other changes are part of this PR?

jeffvg:
overall the page does look different. Dev is on left, spinner width branch on right

image

ektasawant:
> I'm taking a look at this now but initial impressions are that there's a lot of code changes for such a small issue. What other changes are part of this PR?

While working on react design pattern (spinner overlay) got to know that ng design pattern is not having following specs as per design.
1. Width should be as per design.
2. Header and footer should not scroll only card content is scrollable.
3. In mobile remove padding from body

This PR contains changes for fixing above issues. I have updated the same her  [104](https://github.com/pxblue/angular-design-patterns/issues/104)

jeffvg:
do we need to revisit the Angular spinner example for this?
Also been seeing this more often.
should we switch to use this for react?
image

image

emclaug2 (commented):
Could you adjust the margin on the checkbox & 'I Agree' text?   It's a little tight. 
Screen Shot 2021-09-14 at 10 50 11 AM

emclaug2 (approved):
Looks good to me, just see comment from before.

jeffvg:
> do we need to revisit the Angular spinner example for this?
> Also been seeing this more often.
> should we switch to use this for react?
> image
> 
> image

same here all looks good except for the question if we should try out the "styleFunctionSx"

emclaug2:
One last thing I noticed; can you vertical center the content? 

image

ektasawant:
> styleFunctionSx



> do we need to revisit the Angular spinner example for this?
> Also been seeing this more often.
> should we switch to use this for react?
> image
> 
> image



> same here all looks good except for the question if we should try out the "styleFunctionSx"
I think this issue is in react design pattern repository.

jeffvg (approved):

ektasawant:
> One last thing I noticed; can you vertical center the content?
> 
> image

Fixed please review

ektasawant:
> Could you adjust the margin on the checkbox & 'I Agree' text? It's a little tight.
> Screen Shot 2021-09-14 at 10 50 11 AM

fixed please review.

emclaug2 (approved):

PR #131, opened 2021-09-13
Feature/forms validation tests

jeffvg:
- add form validation cypress tests
- update view port test
- 

#### To Test:
- yarn && yarn start
- open new terminal and yarn cy:run


emclaug2 (commented):
@jeffvg Can you resolve the merge conflicts for this PR?  Tests run locally just fine, I'm looking through them now.

cypress/integration/form-validation/in-list.spec.js
```diff
@@ -0,0 +1,25 @@
...
+
+    it('should display page title', () => {
+        cy.get('[data-cy=pxb-toolbar]').should('contain', 'In a List');
+    });
+
+    it('should display ip address left aligned', () => {
+        cy.get('[data-cy=ip-address]').invoke('css', 'text-align').should('equal', 'left')
+    });
+
+    it('should display disabled toggle in #424e54 opacity 0.3', () => {
```
emclaug2, line 17: There's no 'disabled' switch as part of this example; should this be "should display unselected toggle" ?

cypress/integration/form-validation/in-list.spec.js
```diff
@@ -0,0 +1,25 @@
...
+        cy.get('.mat-slide-toggle-bar').invoke('css', 'background-color').should('equal', 'rgba(66, 78, 84, 0.3)')
+    });
+
+    it('should display enabled toggle in #0088f2 opacity 0.5 ', () => {
+        cy.get('.mat-slide-toggle-bar').click()
+        .invoke('css', 'background-color').should('equal', 'rgba(0, 136, 242, 0.5)')
```
emclaug2, line 23: I'm not sure why we're doing theme enforcement checks as part of these unit tests.  Is this to increase the code coverage?  If we ever add a dark theme or make theme changes then these tests may fail in the future.

emclaug2 (approved):
Tests look good & pass locally; I think this is pretty thorough.

I'm not the biggest fan of testing colors to determine if a form field is in an error state though. 
 I'd rather test for whether a `mat-form-field` has the conditional classes that make it an error state.   Down the road testing colors may cause some issues if we ever support light/dark theme, or if our theme colors get tweaked.

huayunh:
> I'm not the biggest fan of testing colors to determine if a form field is in an error state though.

+1. This might cause a lot of extra testing works in the future.

cypress/integration/form-validation/in-list.spec.js
jeffvg: yes just to see how it's handled in cypress and reporting. Next quarter we might use cypress mount component for this stuff once the plug-in is flushed out

cypress/integration/form-validation/in-list.spec.js
jeffvg: changed this to unselected and the other as selected

jeffvg:
@emclaug2 @huayunh I'm going to update the tests in 4 of the design patterns to verify state rather than color displayed

jeffvg:
this should be good now. Removed the asserts on colors and checked toggle state on list example

PR #132, opened 2021-09-13
Feature/collapsible app bar

emclaug2:
-  Add collapsible appbar example


emclaug2:
@jeffvg This example is replacing the existing collapsible appbar example so I'm going to try to update the cypress tests.

jeffvg:
Running in to this error checking the position on collapsible appbar.
(only within the route http://localhost:4200/app-bar/...)
- scroll to collapse the app-bar
- navigate to search bar
- navigate back to collapsible appbar

When navigating back we don't do a reload/refresh of the page

image

emclaug2:
@jeffvg This issue should be fixed now.

You should be able to observe that when you go to the Search page, scroll down, then go the Collapsible AppBar page, the animation from Expanded to Collapsed happens immediately.  Does this seem like a bug?  I'd imagine the appbar should just initialize in it's collapsed state instead of immediately transitioning to it.

jeffvg:
> @jeffvg This issue should be fixed now.
> 
> You should be able to observe that when you go to the Search page, scroll down, then go the Collapsible AppBar page, the animation from Expanded to Collapsed happens immediately. Does this seem like a bug? I'd imagine the appbar should just initialize in the collapsed state instead of immediately transitioning to it.

I see what you mean. Technically in a app this would be a bug but I don't think it's a huge deal for what we are doing. It would be nice to have the 2 pages in the same route treated separately (not scrolling with each other) but not sure if that's possible without a lot of code changes

jeffvg (approved):
looks good

emclaug2:
I talked to Joe and ideally whenever a route change occurs, the scroll distance should be reset so the user is placed at the top of the page.

huayunh:
> whenever a route change occurs, the scroll distance should be reset

Design-wise, this depends on where the user comes from. If a user traces back from a subpage to its root, sometimes you would want to preserve that scroll distance. For this project, it is perfectly fine to go back to the top.

joebochill:
Huayun is right. I think the browser default behavior is usually to scroll to top when you go forward to a new page, and restore the scroll position when using back/forward. I think SPAs kind of screw around with that though.

emclaug2:
For a quick fix I could add a route listener that manually scrolls a user to the top on route changes; it won't preserve previous scroll distance if a user hits the back button though.

huayunh (approved):
Looks good.

emclaug2:
I'll make a separate PR for the scroll-top functionality since it's going to affect all pages in the showcase.

PR #133, opened 2021-09-14
Feature/dropdown toolbar

emclaug2:
- Add dropdown toolbar appbar example. 
- Has a menu overlay on desktop view, bottomsheet view on mobile.


jeffvg (approved):
looks good

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
```diff
@@ -0,0 +1,17 @@
+<pxb-dropdown-toolbar #toolbar title="Title" [subtitle]="getSelected() || 'subtitle'">
```
huayunh, line 1: may want to give it a z4 dropdown shadow

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
```diff
@@ -0,0 +1,17 @@
...
+        *ngIf="isSmall"
+        style="margin-left: -8px"
+        mat-icon-button
+        (click)="toggleMenu()"
+    >
+        <mat-icon>menu</mat-icon>
+    </button>
+    <ng-container pxb-toolbar-menu>
+        <ng-container *ngFor="let item of items">
+            <button mat-menu-item (click)="setSelected(item)">{{ item }}</button></ng-container
```
huayunh, line 14: Can we use list items instead of buttons here? Like what you had in `app-dropdown-toolbar-bottomsheet`

huayunh (commented):
Probably doesn't matter, just an observation that if you resize to mobile, toggle bottom sheet, then go back to desktop size, you will see that bottom sheet is still there.

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
emclaug2: You got it.

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
emclaug2: I based this example off of what we're doing in the storybook demo.  Just to sanity check, ng material's menu demos also use buttons with the `mat-menu-item` directive applied.

emclaug2:
> Probably doesn't matter, just an observation that if you resize to mobile, toggle bottom sheet, then go back to desktop size, you will see that bottom sheet is still there.

I think this behavior will eventually be baked into the Appbar component itself once the DropdownToolabr is deprecated so I'd really prefer not to fix it here, yet.  If that's fine I can make an issue for this so we don't forget about it.

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
huayunh: The design side has been using the more listy look.

(I personally don't mind, but just want to stay consistent)

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.ts
```diff
@@ -0,0 +1,65 @@
...
+    constructor(
+        private readonly _drawerService: StateService,
+        private readonly _breakpointObserver: BreakpointObserver,
+        private readonly _bottomSheet: MatBottomSheet,
+        private readonly _stateService: DropdownToolbarStateService
+    ) {}
+
+    ngOnInit(): void {
+        this._breakpointObserver
+            .observe([Breakpoints.Small, Breakpoints.Handset])
```
huayunh, line 29: Can we use `XSmall` here? `Small` is still a bit wide for a bottom sheet.

emclaug2: You got it, this is using xsmall now.

src/app/pages/app-bar/dropdown-toolbar/dropdown-toolbar.component.html
emclaug2: I'm still a little confused what makes the menu overlay look listy vs using buttons.

emclaug2: Bottomsheet is updated to look more like a list; for the toolbar's menu, it has some top & bottom padding.  If this should be changed by default, then this sounds like a component-library change. 

image

huayunh (approved):

PR #134, opened 2021-09-14
Feature/gloabl search

emclaug2:
- Global Search example


jeffvg (approved):
seems good to me

huayunh (changes requested):
# 1
image

# 2
Can we also remove one of the classes `mat-elevation-z2` or `mat-elevation-z4` from the toolbar?

# 3
Some how the text feels off-centered to me. But when I fiddled around with the CSS I couldn't get it fixed either.
image
Also, somehow the font family is overridden by my browser, and it is currently Arial instead of Open Sans. Is it the same way on your side?

emclaug2:
@huayunh The demo is showing Open Sans for me, that's strange; does that only happen for this feature branch? 

Also for the SearchBar input field; can you tell if the input is off-centered or is the icons?

All the others points have been fixed so it's ready for another round of review.

huayunh:
Did a fresh clone and ran localhost in incognito, somehow it is still not open sans... 

Here is what it looked like in Chrome, rendered as `Arial`
image

And here is what it looked like in Firefox, rendered as `-apple-system`
image

Will log a bug for this.

----

The icon appears like 1px down the center line while the text "Search" is like 2px above. I couldn't figure out why, but this is out of the control for the current project. Just bugging me.

huayunh:
When you clear the search text, can you also auto-focus the search field?

emclaug2:
Ok @huayunh  I think I fixed the font-family issue on the Input field & the input field clear button focuses the input now.

huayunh:
On mobile, the hamburger menu icon is gone

image

Everything else looks good

emclaug2:
@huayunh The menu icon should now be appearing on mobile 👍

huayunh (approved):
code change looks good

PR #136, opened 2021-09-16
Feature/page wide search

emclaug2:
- Page Wide Search


huayunh:
Hmm, if you first go to a page with white background, say, http://localhost:4200/forms-and-validation/in-a-list, and then you navigate to page-wide search, you would see that the background stays white. But if you go to the Page-Wide Search page directly from the landing page, it renders OK.

image

huayunh (commented):
image
This should not be an icon button. 

image
Can we use the standard / underlined form field variant for this guy?

Functionality-wise everything looks nice

emclaug2:
> Hmm, if you first go to a page with white background, say, http://localhost:4200/forms-and-validation/in-a-list, and then you navigate to page-wide search, you would see that the background stays white. But if you go to the Page-Wide Search page directly from the landing page, it renders OK.

We've got another page's styles bleeding into this one; I'll figure out where it's coming from & fix it.

emclaug2:
> We've got another page's styles bleeding into this one; I'll figure out where it's coming from & fix it.

There's quite a bit of style bleeds from the form examples which will require that I edit a large amount of files.  Sounds to me a like a separate PR just to fix the style bleeding.

emclaug2:
#138  opened to fixed style bleeds.

emclaug2:
> Can we use the standard / underlined form field variant for this guy?

Should be good now & ready for another round of review @huayunh

src/app/app-routing.module.ts
```diff
@@ -56,6 +57,11 @@ export const ROUTES: DrawerItem[] = [
                 path: 'search',
```
huayunh, line 57: Let's call this "global search" to differentiate

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
```diff
@@ -0,0 +1,70 @@
...
+        <div class="mat-title" [style.marginLeft.px]="isSmall ? 24 : 0">Page Search</div>
+        <pxb-spacer></pxb-spacer>
+
+        <div pxb-actions class="appbar-actions">
+            <button mat-icon-button>
+                <mat-icon style="margin: 0 8px">help</mat-icon>
+            </button>
+            <button mat-icon-button>
+                <mat-icon style="margin: 0 8px" [matBadge]="88" matBadgeColor="warn">notifications</mat-icon>
+            </button>
```
huayunh, line 15: `0 4px` margin works better for these icons

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
```diff
@@ -0,0 +1,70 @@
+<mat-toolbar data-cy="pxb-toolbar" color="primary" class="main-header mat-elevation-z4 t" [class.hidden]="toggleSearch">
```
huayunh, line 1: what does this `t` class do?

huayunh: And it looks like this shadow has gone missing
image

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
```diff
@@ -0,0 +1,70 @@
...
+</mat-toolbar>
+<mat-divider *ngIf="isSmall"></mat-divider>
+
+<div class="content-body-container" [class.small]="isSmall">
+    <div class="content-body" [class.small]="isSmall">
+        <div *ngIf="!isSmall" class="desktop-search">
+            <pxb-spacer></pxb-spacer>
+            <mat-form-field appearance="standard" style="width: 160px">
+                <mat-icon matPrefix class="desktop-search-icon">search</mat-icon>
+                <input matInput type="text" placeholder="Search Asset" [(ngModel)]="searchText" />
```
huayunh, line 53: Make your desktop and mobile text match each other

```suggestion
                <input matInput type="text" placeholder="Search Assets..." [(ngModel)]="searchText" />
```

huayunh (approved):

jeffvg:
did we want to include 'X' to clear search like what was done with the react example? @huayunh 

image

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
emclaug2: Ok you got it.

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
emclaug2: I've re-added the elevation here.

src/app/pages/app-bar/page-wide-search/page-wide-search.component.html
emclaug2: Fixed 👍

huayunh:
when the text field isn't empty, there shall be an X, I think.

emclaug2:
> when the text field isn't empty, there shall be an X, I think.

Yep and this is how this example is handling it; whenever the X is hit the search is cleared.

PR #149, opened 2021-10-04
update broken links, add new pattern links

ektasawant:
- Fix design pattern links in readme
- Add new design pattern description


joebochill (approved):

joebochill:
At some point we may want to break these into multiple tables for each of the top level sections ("In a list" doesn't give a whole lot of context without knowing that it's underneath Form Validation)

PR #157, opened 2021-11-03
Feature/pxblue 2551 tree structure

emclaug2:
- Tree structure example

Can be seen here: https://pxblue-angular-design-patterns.web.app/lists/tree-structure

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -30,6 +30,7 @@ import { MenuInADrawerComponent } from './pages/account-menu/in-a-drawer/menu-in
 import { DropdownToolbarComponent } from './pages/app-bar/dropdown-toolbar/dropdown-toolbar.component';
 import { SearchBarComponent } from './pages/app-bar/search-bar/search-bar.component';
 import { ContextualAppBarComponent } from './pages/app-bar/contextual/contextual.component';
+import { TreeListComponent } from './pages/list/tree/tree.component';
 
 export type DrawerItem = {
     title: string;
@@ -166,6 +167,11 @@ export const ROUTES: DrawerItem[] = [
                 title: 'Responsive Table',
                 path: 'responsive-table',
                 component: ResponsiveTableComponent,
+            },
+            {
+                title: 'Tree Structure',
+                path: 'tree-structure',
+                component: TreeListComponent,
             },
         ],
     },
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -54,6 +54,7 @@ import { MenuInDrawerModule } from './pages/account-menu/in-a-drawer/in-a-drawer
 import { DropdownToolbarModule } from './pages/app-bar/dropdown-toolbar/dropdown-toolbar.module';
 import { ContextualAppbarModule } from './pages/app-bar/contextual/contextual.module';
 import { SearchBarModule } from './pages/app-bar/search-bar/search-bar.module';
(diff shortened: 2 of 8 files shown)


emclaug2:
@mvbela there's some stuff that feels off when clicking around in the example.  

1. For items that don't have any children, will clicking on the row select the item?
2. Is selection hierarchical?  If I select a parent, are the children also selected?
3. For items with no children, does clicking the row select it?

huayunh:
Can you make these arrows centered? 
image

And you may want to work on responsiveness... I don't know if they are intended to be centered this way on mobile.

https://user-images.githubusercontent.com/8997218/140258779-e0554fc1-75cf-4202-ac6c-5561e67fbb32.mov

For the desktop view, you would want a 16px padding when the container width is close to the card width. See here: https://github.com/pxblue/react-design-patterns/pull/101#issuecomment-943667444

mvbela:
Hi @emclaug2 ,

1. The items are only selected when the radio button is clicked, so it should not select the item if the user clicks the row.
2. Yes, if you select the parent, the children are also selected. This means that for the user to select an isolated child, (s)he should click the keyboard_arrow icon to open the parent and click on the item.

Will update the Figma file with the these comments!

emclaug2:
@mvbela Thanks for the clarification!  One more follow-up question here; does the design have to use radio-buttons?  I'm in a situation right now where I'm fighting the framework & struggling to unselect a radio button once selected.  Checkboxes would be easier to implement but only if the design works.

emclaug2:
This example has been updated to use the original radio buttons; should be ready for review now. 
Check it out here: https://pxblue-angular-design-patterns.web.app/lists/tree-structure

@huayunh For the off-centered chevrons, this is something that's going to affect all the mat expansion panels.  I'm trying to figure out where the best place for this fix should go.

mvbela:
Hi @emclaug2,
I've noticed some details on the tree structure. 
 
**On desktop version:**

1. There should be rounded corners (4px)

Captura de ecrã 2021-11-09, às 15 47 03


**Both in mobile and desktop:**

2. Sometimes it happened that I could select multiple radio buttons

Captura de ecrã 2021-11-09, às 15 38 03

3. When an item is selected, the other items do not show change of color in hover

4. There should be a divider between all items

Captura de ecrã 2021-11-09, às 17 05 25

5. On selected items, there shouldn't be a shadow, but a divider

Captura de ecrã 2021-11-09, às 17 08 16

emclaug2:
Thanks for the review @mvbela ; should be ready for another look! I was unable to reproduce #2 though; you're still able to select more than 1 item please let me know the steps to reproduce.

daileytj:
The "move file" button is disabled when there is no folder selected on the mobile view, should this functionality be matched in the desktop view?
### mobile
Screen Shot 2021-11-10 at 9 07 24 AM
### desktop
Screen Shot 2021-11-10 at 9 07 10 AM

daileytj:
This is more of a design comment/critique, but why do we have active "add folder" and "move folder" buttons that don't do anything while also not letting the user know that this is a placeholder. If I saw this example and needed the functionality I would expect to be able to just copy/paste the code to get all the move/add folder functionality.

mvbela:
Yes, the "Move File" button should also be disabled when there is no item selected on desktop

emclaug2:
Demo has been updated.

daileytj (approved):

PR #170, opened 2021-12-28
Feature/2547 ng global list button panel

ektasawant:
Fixes 2547 .

- Desktop view
- Mobile view

#### Screenshots:

Screenshot 2021-12-28 at 5 06 26 PM
Screenshot 2021-12-28 at 5 06 19 PM
Screenshot 2021-12-28 at 5 06 04 PM

Screenshot 2021-12-28 at 5 05 40 PM

#### To Test:
- http://localhost:4200/lists/in-button-panel


daileytj:
- [ ] Empty state icon should be warning_amber (If its not available in the latest icon set then the current one is ok)
- [ ] Can we make the modal width 280px to match the designs
- [ ] Should the modal be centered in the content window on desktop like the design? Or are we cool with it just being centered within the entire app view?
- [ ] I prefer the vertically centered empty state that is built out, but the designs do not have this styling. @mvbela && @huayunh, is there a standard for how we want to display empty states in a page view?
- [ ] In mobile view, when there are no items, the delete icon button should not appear
Screen Shot 2021-12-28 at 12 20 05 PM

daileytj (approved):

Issue #175, opened 2021-10-18
mat-expansion-panel-header divider color should be the same as other dividers


huayunh:
## Describe the desired behavior

`<mat-expansion-panel-header>` should be like other `.pxb .mat-divider`, which has the style 

```css
border-top-color: rgba(66, 78, 84, 0.12);
```

## Describe the current behavior

`<mat-expansion-panel-header>` currently has a style `border-bottom: 1px solid rgba(0, 0, 0, 0.12);`

Here's what it looks like in the context

![image](https://user-images.githubusercontent.com/8997218/137811405-4aaa41b4-a2d6-42d9-a30c-d1634ba75b35.png)

PR #245, opened 2022-04-06
mat-expansion-panel-header border match with divider color

ektasawant:
Fixes [175](brightlayer-ui/angular-themes#107) .

- fix mat expansion header border for status list example
- fix mat expansion header border for inline actions example
- 

#### Screenshots:
Screenshot 2022-04-06 at 5 22 08 PM
Screenshot 2022-04-06 at 5 22 24 PM

#### To Test:
- http://localhost:4200/lists/status-list
- http://localhost:4200/lists/inline-local-actions

inspect mat-expansion-panel-header and observe `border-bottom` css property .

### DIFF
--- a/src/app/pages/list/status-list/status-list.component.scss
+++ b/src/app/pages/list/status-list/status-list.component.scss
@@ -45,7 +45,7 @@ section.list-section {
                     margin-top: -8px;
                 }
                 &.mat-expanded {
-                    border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+                    border-bottom: 1px solid rgba(map-get($blui-black, 500), 0.12);
                     .mat-expansion-indicator {
                         margin-top: 4px;
                     }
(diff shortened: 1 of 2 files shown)


src/app/pages/list/status-list/status-list.component.scss
```diff
@@ -45,7 +45,7 @@ section.list-section {
                     margin-top: -8px;
                 }
                 &.mat-expanded {
-                    border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+                    border-bottom: 1px solid rgba(map-get($blui-black, 500), 0.12);
```
emclaug2, line 48: This seems like a fix that should be applied at the theme-level instead of the design patterns.
Whenever the issue #175 was made, the requested changes don't only apply to the examples within the design patterns repo.  All expansion panels should have their light/dark theme border-bottoms updated.

ektasawant: Sure @emclaug2 , I have transferred this issue. The reason I fixed this issue in NG design pattern is here https://github.com/brightlayer-ui/angular-themes/issues/107#issuecomment-1091135279. But right now I have tried this changes it is not effecting drawer. So, fixed issue in angular themes. Thanks!!!

ektasawant: Closing this PR. Since issue is transferred to angular-themes

